Issue #73, opened 2025-09-25
Pass as many Errors as Possible to the Client + Panic on Violated Invariants


0xNeshi:

0xNeshi:
Already handled in other PRs. Future PRs will be ensured to comply with this issue's requirement

PR #74, opened 2025-09-25
Reorg live mode

### LINKED ISSUE #57: Live Mode Reorg Handling

LeoPatOZ:
Resolves #57 

After a lot of research _and headache_ , it seems that block stream you get from `eth_subscribe` actually returns the latest head from the reorg point. 
See:
https://docs.metamask.io/services/reference/ethereum/json-rpc-methods/subscription-methods/eth_subscribe#parameters

And confirmed from personal testing. 

The main source of confusion is that querying the latest head will not get you latest reorged head (event though it may have been reorged) i.e. Head = 10, Reorg Block = 8. Get latest block will always return 10 but the subscription will go back to 8. This means we can simply check if the current block is the more than the one returned by the subscription. 

Still unclear:
1. Is this always safe.... (should we sill implement parent hash checks). Im worried about a potentially rare edge case where latest block is 10, 8 is reorg, subscription fails for some reason / skips a block, 9 is returned. 
2. Should we care about user inputted reorg depth?
3. How to send this error to event scanner without stopping it

Issue tracked https://github.com/OpenZeppelin/Event-Scanner/issues/76

Need to still solve https://github.com/OpenZeppelin/Event-Scanner/issues/73


LeoPatOZ:
> > The main source of confusion is that querying the latest head will not get you latest reorged head (event though it may have been reorged) i.e. Head = 10, Reorg Block = 8. Get latest block will always return 10 but the subscription will go back to 8.
> 
> So the chain would be:
> 
> * pre-reorg: ...- 8 - 9 - 10
>   (reorg happens)
> * post-reorg: ...- 8 - 9 - 10 (same numbers, different hashes)
> 
> But the subscription would return this stream:
> 
> * ...
> * 8
> * 9
> * 10
>   (REORG)
> * 8
> * 9
> * 10
> * ...
> 
> @LeoPatOZ can you confirm?
> 
> > This means we can simply check if the current block is the less than the one returned by the subscription.

Yes i confirm (this is how Ive done it now)

> Shouldn't the check be whether the last processed block ("current") is _larger_ than the newly received block (because subscription returns earlier blocks post-reorg)?

Sorry yes i actually did that in my code just confused it in the PR description


> Also, we'll definitely need this check if we want to be able to notify the calling code that a reorg happened by returning an error prior to returning new blocks.

Yeah currently theres an issue with how the event scanner is impl but i agree

> > Still unclear:
> > 
> > 1. Is this always safe.... (should we sill implement parent hash checks). Im worried about a potentially rare edge case where latest block is 10, 8 is reorg, subscription fails for some reason / skips a block, 9 is returned.
> > 2. Should we care about user inputted reorg depth?

> > 3. How to send this error to event scanner without stopping it
> 
> 1. Let's imagine something like this can happen. How would we determine that this edge case happened? Do you have an algorithm in mind (even if imperfect)?

I dont think its actually possible, because you have no way to checking if the prev hash has changed since we arent storing it. Let say

8 -> 9 -> 10 -> 11
9 is reorged
susbcirption for some reason fails and decides to return 10.
We have no way of checking if 10 is the latest reorged block, as we didn't save 9's original hash. If we check is 10 parent hash == 9 hash, then that will be true since it will be checking 9's new hash. 
My only solution is to go back a couple blocks just to be extra safe, so return reorg block - reorg depth --> latest block


> 2. Yes. I asked David specifically about this during our last call, and the idea is to give higher priority to user's reorg depth and fallback to our default value and/or internal logic otherwise. Though I think he was talking about historical mode specifically, where this does make more sense. In live mode, I don't see why we would even need to depend on the reorg depth value, since the subscription itself will presumably give us the exact new post-reorg block (so no need to calculate); this assumes 1) is handled of course.
> 3. I think this should be straightforward:
>    
>    1. Await new block header from subscription
>    2. Determine if reorg happened (e.g. block number comparison, or custom algorithm from 1.)
>    3. Invoke `block_sender.send(Err(Error:ReorgOccurred)).await?`
>    4. Immediately invoke `event_sender.send(Ok(post_reorg_block..=post_reorg_block)).await?`
>    5. Update `current` (we should really rename this variable, it's confusing)
>    6. Go to i.

Yup this is how ive done it. A question is should we just the local current variable or the global one?

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+
+#[tokio::test]
+async fn reorg_rescans_events_with_rewind_depth() -> anyhow::Result<()> {
+    let anvil = spawn_anvil(0.1).unwrap();
+    let provider = build_provider(&anvil).await.unwrap();
+
+    let contract = deploy_counter(provider.clone()).await?;
+    let contract_address = *contract.address();
+    let blocks = Arc::new(Mutex::new(Vec::new()));
+    let callback = Arc::new(BlockOrderingCallback { blocks: Arc::clone(&blocks) });
```
0xNeshi, line 30: Update your branch, we no longer have callbacks

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+    let reorg_options = ReorgOptions { depth: reorg_depth, tx_block_pairs };
+
+    provider.anvil_reorg(reorg_options).await.unwrap();
+    let new_latest_block =
+        provider.get_block_by_number(BlockNumberOrTag::Latest).full().await?.unwrap();
+    // sanity checks, block numb stays the same but hash changes
+    assert_eq!(new_latest_block.header.number, latest_block.header.number);
+    assert_ne!(new_latest_block.header.hash, latest_block.header.hash);
+    let new_block = provider
```
0xNeshi, line 83:
```suggestion

    let new_block = provider
```
nit: add new line, increase clarity

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let mut scanner = EventScannerBuilder::new()
+        .with_event_filter(filter)
+        .with_reorg_rewind_depth(6)
+        .connect_ws::<Ethereum>(anvil.ws_endpoint_url())
+        .await?;
+
+    tokio::spawn(async move { scanner.start(BlockNumberOrTag::Latest, None).await });
+
+    let mut expected_event_block_number = vec![];
```
0xNeshi, line 45:
```suggestion
    let mut expected_event_block_numbers = vec![];
```
implies vector type

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+    let blocks_clone = Arc::clone(&blocks);
+    let event_counting = async move {
+        while blocks_clone.lock().await.len() < initial_events {
+            sleep(Duration::from_millis(100)).await;
+        }
+    };
+
+    if timeout(Duration::from_secs(1), event_counting).await.is_err() {
+        bail!("expected {initial_events} events, got {}", blocks.lock().await.len());
+    }
```
0xNeshi, line 62: nit: imo this check is redundant and just adds test exec. time, other tests confirm this works, so this one can assume it does, too, and focus on just testing reorg logic

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let mut tx_block_pairs = vec![];
+    let num_new_events = 3;
+    let latest_block = provider.get_block_by_number(BlockNumberOrTag::Latest).await?.unwrap();
+    let reorg_depth = 10;
+    for _ in 0..num_new_events {
+        let tx = contract.increase().into_transaction_request();
+        // 0 is offset to reorg depth
+        tx_block_pairs.push((TransactionData::JSON(tx), 0));
```
0xNeshi, line 71: Should all of the re-executed transactions have the same offset?

Looking at how Anvil works (see [anvil_reorg source](https://github.com/foundry-rs/foundry/blob/f8dd6aef21a7a02a6b7471ee06427f155c03a3b3/crates/anvil/src/eth/api.rs#L2351)), the [original txs in this test](https://github.com/OpenZeppelin/Event-Scanner/pull/74/files#diff-03bbf17c3bbf94e2875ec73dbd60b3a44a65f4e3fff908ece7a6b8b9516d16efR49) might have been created each in a separate block, doesn't Anvil work like that in this test?
If so, the realistic behavior here would be execute each transaction in the same post-reorg block, meaning the offset should be increased appropriately

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+        // pushing to latest block - roerg depth + 1;
```
0xNeshi, line 72:
```suggestion
```
nit: redundant, code is self-explanatory

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    if timeout(Duration::from_secs(1), event_counting).await.is_err() {
+        bail!("expected {initial_events} events, got {}", blocks.lock().await.len());
+    }
+
+    let mut tx_block_pairs = vec![];
+    let num_new_events = 3;
+    let latest_block = provider.get_block_by_number(BlockNumberOrTag::Latest).await?.unwrap();
+    let reorg_depth = 10;
```
0xNeshi, line 67: Are we 100% that Anvil will always have at least 10 blocks at this point? 
Since the test executed 5 txs before this, it is safer to assume that it will always have at least 5 blocks, isn't it?

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+    for _ in 0..num_new_events {
+        let tx = contract.increase().into_transaction_request();
+        // 0 is offset to reorg depth
+        tx_block_pairs.push((TransactionData::JSON(tx), 0));
+        // pushing to latest block - roerg depth + 1;
+        expected_event_block_number.push(latest_block.header.number - reorg_depth + 1);
```
0xNeshi, line 73: Because all txs post-reorg are included in the same block, this will push the same value `num_new_events` times

tests/live_mode/reorg.rs
```diff
@@ -0,0 +1,116 @@
...
+            "Post-reorg events not rescanned in time. Expected at least {}, got {}",
+            initial_events + 4,
+            current_len
+        );
+    }
+    let final_blocks: Vec<_> = blocks.lock().await.clone();
+    assert!(final_blocks.len() == initial_events + num_new_events);
+    assert_eq!(final_blocks, expected_event_block_number);
+    // sanity check that the block number after the reorg is smaller than the previous block
+    assert!(final_blocks[initial_events] < final_blocks[initial_events - 1]);
```
0xNeshi, line 113: Not sure what this is checking, can you elaborate?

0xNeshi (commented):
Second pass

tests/live_mode/reorg.rs
LeoPatOZ: removed!

tests/live_mode/reorg.rs
LeoPatOZ: I dont think theres a correct answer, in a real reorg you could imagine that all events were removed from subsequent blocks and instead emitted in the same block, or that they were emitted in subsequent blocks. 

I just decided to emit them all in the same block, i dont see a strong reason to change it but can do :)

tests/live_mode/reorg.rs
LeoPatOZ: done!

src/block_range_scanner.rs
```diff
@@ -561,23 +568,43 @@ impl<N: Network> Service<N> {
...
                 let mut stream = ws_stream.into_stream().skip_while(|header| header.number() < cur);
-                while let Some(header_resp) = stream.next().await {
-                    info!(block_number = header_resp.number(), "Received block header");
-
-                    if sender.send(Ok(current..=header_resp.number())).await.is_err() {
+                while let Some(incoming_block) = stream.next().await {
+                    let incoming_block_num = incoming_block.number();
+                    info!(block_number = incoming_block_num, "Received block header");
+
+                    if incoming_block_num < current {
```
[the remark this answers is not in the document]
LeoPatOZ: okay changed it!

src/block_range_scanner.rs
```diff
@@ -561,23 +568,43 @@ impl<N: Network> Service<N> {
...
+                        // TODO: send reorg err - issue is this causes event scanner to stop
+                        // if sender.send(Err(Error::ReorgDetected)).await.is_err() {
+                        //     warn!("Downstream channel closed, stopping live blocks task");
+                        //     return;
+                        // }
```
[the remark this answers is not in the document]
LeoPatOZ: okay i did that instead - too much hassle to deal with it i this PR

src/block_range_scanner.rs
```diff
@@ -561,23 +568,43 @@ impl<N: Network> Service<N> {
...
+                        warn!("Reorg detected: sending forked range");
+                        // TODO: should we send the incoming block range or incoming block num -
+                        // reorg depth? The incoming block should be the
+                        // latest block from the reorg point so no real need
+                        // tbd
+                        if sender
+                            .send(Ok(incoming_block_num..=incoming_block_num + 1))
```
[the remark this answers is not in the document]
LeoPatOZ: solved now `incoming_block_num..=incoming_block_num` no need for the + 1

tests/live_mode/reorg.rs
LeoPatOZ: removed now dont really need it tbh

src/block_range_scanner.rs
```diff
@@ -558,26 +565,41 @@ impl<N: Network> Service<N> {
...
+
+                    if incoming_block_num < expected_next_block {
+                        if sender.send(Err(Error::ReorgDetected)).await.is_err() {
+                            warn!("Downstream channel closed, stopping live blocks task");
+                            return;
+                        }
+                        warn!("Reorg detected: sending forked range");
+                        // TODO: should we send the incoming block range or incoming block num -
+                        // reorg depth? The incoming block should be the
+                        // latest block from the reorg point so no real need
```
0xNeshi, line 591: Does this comment still make sense?

src/block_range_scanner.rs
```diff
@@ -558,26 +565,41 @@ impl<N: Network> Service<N> {
...
+                while let Some(incoming_block) = stream.next().await {
+                    let incoming_block_num = incoming_block.number();
+                    info!(block_number = incoming_block_num, "Received block header");
+
+                    if incoming_block_num < expected_next_block {
+                        if sender.send(Err(Error::ReorgDetected)).await.is_err() {
+                            warn!("Downstream channel closed, stopping live blocks task");
+                            return;
+                        }
+                        warn!("Reorg detected: sending forked range");
```
0xNeshi, line 588: Move this above the if check

Issue #64, opened 2025-09-24
Stream Only Events That Have the Specified Number Block Confirmations


0xNeshi:
This configuration should be applied per event filter.

Modes:
- [ ] Live - https://github.com/OpenZeppelin/Event-Scanner/pull/98
- [x] ~~Historical~~ EDIT: does not apply after all
- [ ] Sync - https://github.com/OpenZeppelin/Event-Scanner/pull/98

Issue #58, opened 2025-09-23
Sync Mode Reorg Handling


0xNeshi:
Sync mode = historic -> live switch

0xNeshi:
@LeoPatOZ can you assign yourself to the issue for future reference?

PR #98, opened 2025-09-30
Add block confirmations for live + sync mode

LeoPatOZ:
Resolves https://github.com/OpenZeppelin/Event-Scanner/issues/64
Resolves https://github.com/OpenZeppelin/Event-Scanner/issues/58

### DIFF
--- a/src/block_range_scanner.rs
+++ b/src/block_range_scanner.rs
@@ -1198,6 +1198,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "Flaky test, see: https://github.com/OpenZeppelin/Event-Scanner/issues/109"]
     async fn continuous_blocks_if_reorg_less_than_block_confirmation() -> anyhow::Result<()> {
         let anvil = Anvil::new().try_spawn()?;
 
@@ -1242,6 +1243,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "Flaky test, see: https://github.com/OpenZeppelin/Event-Scanner/issues/109"]
     async fn shallow_block_confirmation_does_not_mitigate_reorg() -> anyhow::Result<()> {
         let anvil = Anvil::new().block_time(1).try_spawn()?;
 


0xNeshi (commented):
Good job! Left some comments

src/block_range_scanner.rs
```diff
@@ -1102,6 +1116,51 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn live_mode_respects_block_confirmations() -> anyhow::Result<()> {
```
0xNeshi, line 1120: Add a test case for sync mode, too

tests/live_mode/reorg.rs
```diff
@@ -343,11 +343,120 @@ async fn reorg_depth_two() -> anyhow::Result<()> {
...
 
     let final_blocks: Vec<_> = event_block_count.lock().await.clone();
     assert!(final_blocks.len() == initial_events + num_new_events);
     assert_eq!(final_blocks, expected_event_block_numbers);
 
     Ok(())
 }
+
+#[tokio::test]
+async fn block_confirmations_mitigate_reorgs() -> anyhow::Result<()> {
```
0xNeshi, line 356: Add a similar test for sync mode, too.

src/block_range_scanner.rs
```diff
@@ -413,8 +413,10 @@ impl<N: Network> Service<N> {
         };
 
         let reorg_rewind = self.config.reorg_rewind_depth;
+        let block_confirmations = self.config.block_confirmations;
         tokio::spawn(async move {
-            Self::stream_live_blocks(start, provider, sender, reorg_rewind).await;
+            Self::stream_live_blocks(start, provider, sender, reorg_rewind, block_confirmations)
```
0xNeshi, line 418: The current subscription behavior is to expect the next block to be the latest block, so the scanner waits until enough block confirmations occur before returning. This might be unintuitive to the client, who'd expect to immediately start getting blocks that fulfill the block confirmation requirement.

To give an example, imagine the current latest block is 10. If I'm the client and I subscribe to all new blocks which have 5 block confirmations, I would expect to immediately get blocks 5, 6, 7 etc., and not wait for half a minute or more until block 10 has been confirmed 5 times and only then to start streaming.

To address this, we should assign `start - block_confirmations` to the `expected_next_block` parameter, and update the underlying logic accordingly.

0xNeshi (commented):
Found one unintuitive and potentially problematic behavior.

src/block_range_scanner.rs
```diff
@@ -601,16 +611,19 @@ impl<N: Network> Service<N> {
...
+                            .send(BlockRangeMessage::Data(expected_next_block..=confirmed))
+                            .await
+                            .is_err()
+                        {
+                            warn!("Downstream channel closed, stopping live blocks task");
+                            return;
+                        }
 
-                    expected_next_block = incoming_block_num + 1;
+                        expected_next_block = confirmed + 1;
```
0xNeshi, line 625: Add safety comment stating that overflow here cannot realistically happen

src/block_range_scanner.rs
LeoPatOZ: hmm I actually think the current behavior is correct though - if we set `expected_next_block = start - block_confirmations`, then adding more confirmations would change which historical blocks get scanned, which seems wrong for "live scanning."

The confirmation requirement should only affect the reporting delay, not the starting point. The wait is inherent to requiring confirmations on live data. 

Once you waited for the adequate confirmation time there should no longer be a delay.

src/block_range_scanner.rs
```diff
@@ -1102,6 +1116,51 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn live_mode_respects_block_confirmations() -> anyhow::Result<()> {
+        let anvil = Anvil::new().block_time_f64(0.1).try_spawn()?;
+
+        let block_confirmations = 5;
+
+        let client = BlockRangeScanner::new()
+            .with_blocks_read_per_epoch(3)
+            .with_reorg_rewind_depth(5)
```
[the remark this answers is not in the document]
0xNeshi: Will ask for feedback

src/block_range_scanner.rs
```diff
@@ -426,8 +426,10 @@ impl<N: Network> Service<N> {
         };
```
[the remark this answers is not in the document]
0xNeshi: Resolving this and reverting the PR description, we realized that block confirmations make little sense in historic mode.

src/block_range_scanner.rs
0xNeshi: > If current head is 10 and block confirmation is 2, then i wouldn't expect the next block i get to be 8?

But why wouldn't you, though? You don't know what the latest block number is, you just know that you want to subscribe to new blocks that have 2 block confirmations. You're subscribing to the block stream, not just to the latest block.

If you initiate a live subscription with:
- 0 block confirmations - you're expecting to get the latest block immediately
- 1 block confirmations - you're expecting to get the latest block with 1 confirmations immediately
- 2 block confirmations - you're expecting to get the latest block with 2 confirmations immediately
- ...

This is opposed to what you described: "subscribe, but wait until the current block has X confirmations". This would make more sense to me if incorporated into the `StreamFrom` command, where the intention to "stream from a specific block number" is explicit and clear.

That way we'd have:
- `StreamLive` to mean "stream all latest blocks that have X confirmations"
- `StreamFrom` with `start_height = Latest` to mean "stream from the latest block, but wait for X confirmations"

src/block_range_scanner.rs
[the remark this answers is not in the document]
0xNeshi: We don't need this in live mode, only historic and I suspect in catch-up part of the sync mode

src/block_range_scanner.rs
LeoPatOZ: Okay based off our discussion and the poll  -  decided to update this to match the flow described. Also added tests to that effect.

tests/live_mode/reorg.rs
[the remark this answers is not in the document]
LeoPatOZ: okay done!

LeoPatOZ: done!

src/block_range_scanner.rs
```diff
@@ -1102,6 +1116,51 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn live_mode_respects_block_confirmations() -> anyhow::Result<()> {
+        let anvil = Anvil::new().block_time_f64(0.1).try_spawn()?;
+
+        let block_confirmations = 5;
+
+        let client = BlockRangeScanner::new()
+            .with_blocks_read_per_epoch(3)
```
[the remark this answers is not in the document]
LeoPatOZ: fixed in https://github.com/OpenZeppelin/Event-Scanner/pull/108

0xNeshi (approved):

PR #100, opened 2025-09-30
feat: EventFilter Should Accept Multiple Event Signatures

### LINKED ISSUE #48: feat: Allow `EventFilter` to Accept Both Event Signature in Human-readable and Hashed Forms and Multiple Addresses
This is about aligning our assumptions about how `EventFilter` is to be used with how it's actually capable and intuitive to be used.

~~Change to e.g. an enum:~~ EDIT: it's actually more maintainable to track them separately.

~~Depending on the variant, set either [Filter::event](https://docs.rs/alloy-rpc-types-eth/1.0.34/alloy_rpc_types_eth/struct.Filter.html#method.event#method.event) or [Filter::event_signature](https://docs.rs/alloy-rpc-types-eth/1.0.34/alloy_rpc_types_eth/struct.Filter.html#method.event#method.event#method.event_signature)~~ EDIT: turned out there's a better way.

### LINKED ISSUE #99: feat: `EventFilter` Should Accept Multiple Event Signatures
Since the design already permits tracking _specific_ events (if `event == Some(String)`) and _all_ events (if `event == None`),  it makes sense for it to allow tracking _multiple specific_ events. It seems to be more intuitive, too.

0xNeshi:
Resolves #99
Resolves #48


LeoPatOZ (approved):
nice good job!

src/event_filter.rs
```diff
@@ -1,32 +1,52 @@
...
+        let address =
+            self.contract_address.map_or("all contracts".to_string(), |addr| format!("{addr:?}"));
+        let events =
+            if self.events.is_empty() { "all events".to_string() } else { self.events.join(", ") };
+
+        write!(f, "EventFilter(contract: {address}, events: {events})")
+    }
+}
+
+impl Debug for EventFilter {
```
LeoPatOZ, line 42: Is this different from deriving Debug or does this now use the custom display format ?

0xNeshi: Implemented explicitly so that it uses `Display` internally. Added test cases to showcase

0xNeshi:
> @0xNeshi Forgot to mention but since we are breaking the api, should we add this ? #48 or are we not going to do it anymore

The API is technically the same, we're just breaking the previous assumption of `1 event filter = event type` :sweat_smile:  The client can no longer instantiate the struct directly though, that's true.

0xNeshi:
We actually need to add support for multiple contract addresses, too, converting to draft for clarity

LeoPatOZ (approved):
Nice! Good job

PR #108, opened 2025-10-03
API Overhaul

### LINKED ISSUE #49: Determine Default Value for `DEFAULT_REORG_REWIND_DEPTH`
https://github.com/OpenZeppelin/Event-Scanner/blob/e80e7a2f4f44c725e7a0e361f9fa52d6afa9b323/src/block_range_scanner.rs#L103

### LINKED ISSUE #104: feat: Move Appropriate Global Configs Into Commands

### LINKED ISSUE #105: feat: Block Confirmations Should Be Passed Only to Live and Sync Mode Commands

### LINKED ISSUE #107: feat: Blocks Processed per Epoch Config Should Be Passed As Global Config
During live streaming, ~~the scanner will processes 1-2 blocks anyway.~~ EDIT: see edge case that makes this config necessary in live/sync modes as well #111

### LINKED ISSUE #111: feat: Handle Live Mode Edge Case when Block Range is Larger Than Configured Blocks Per Epoch
Edge case: 
- event scanner
- live mode
- subscription provider hangs for whatever reason
- subscription hangs for >2.000 blocks
- subscription continues
- event scanner receives a range of more than 2.000 blocks
- event scanner attempts to fetch logs from the range
- RPC provider errors out due to block range being too large

Conclusion: the "blocks per epoch" config should apply to live/sync modes as well. Maybe rename it to `max_blocks_per_epoch`.

Related #104

### LINKED ISSUE #116: feat: API Overhaul
The more we're developing the scanner, the more it's starting to look like the original design we've had and the more the original design seems appropriate. If we're to do an API overhaul, what better time then during early alpha stages?

Here's a proposal for the API design that solves multiple problems we're trying to tackle at the moment:
- support multiple overlapping streaming modes (live, historic, sync, block range, latest event fetch) 
- clearly differentiate the streaming mode being used while using maintainable code
- clean & easy to use API (dev-friendliness)
- hide complexity (dev doesn't have to instantiate tokio tasks etc., they just configure and use the stream)

Below is a proposal for the API that takes advantage of Rust's type system to ensure the appropriate streaming mode, with actual types returned by builder functions in the comments:

- Live Mode Setup (default):

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .with_block_confirmations(2)                            // LiveEventScanner
    .with_contract_address(contract)                        // LiveEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE)           // LiveEventScanner
    .with_event(MyContract::OtherEvent::SIGNATURE);         // LiveEventScanner

// all configs + filters are set, starts the scanner and type system ensures
// this is now streaming in live mode
let mut event_stream = scanner.stream().await?; 
```

- Sync Mode Setup:

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .from_block(BlockNumberOrTag::Number(1000))             // SyncEventScanner
    .with_block_confirmations(5)                            // SyncEventScanner
    .with_contract_address(contract)                        // SyncEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE);          // SyncEventScanner

// the type system ensures this streams in sync mode
let mut event_stream = scanner.stream().await?;  
```

- Historic Mode Setup:

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .from_block(BlockNumberOrTag::Number(1000))             // SyncEventScanner
    .to_block(BlockNumberOrTag::Number(500))                // HistoricEventScanner
    .with_contract_address(contract)                        // HistoricEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE);          // HistoricEventScanner

// `with_block_confirmations(count: u64)` function does not exist on type `HistoricEventScanner`

// type system ensures this is now streaming in historic mode
let mut event_stream = scanner.stream().await?; 
```

- Fetch Latest X Events:

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .latest(100)                                            // LatestEventScanner
    .with_contract_address(contract)                        // LatestEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE);          // LatestEventScanner

// type system ensures this is now streaming in "fetch latest X events" mode
let mut event_stream = scanner.stream().await?; 
```

- Fetch Latest X Events and Switch to Live:

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .latest(100)                                            // LatestEventScanner
    .then_live()                                            // SyncLatestEventScanner
    .with_contract_address(contract)                        // SyncLatestEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE);          // SyncLatestEventScanner

// type system ensures this is now streaming in "fetch latest X events and then in live" mode
let mut event_stream = scanner.stream().await?; 
```

- Note that we can even combine "Fetch Latest X Events" and block ranges:

```rust
let scanner = EventScanner::new("wss://some.websocket.url") // LiveEventScanner
    .latest(100)                                            // LatestEventScanner
    .from_block(BlockNumberOrTag::Number(1000))             // LatestEventScanner
    .to_block(BlockNumberOrTag::Number(500))                // LatestEventScanner
    .with_contract_address(contract)                        // LatestEventScanner
    .with_event(MyContract::SomeEvent::SIGNATURE);          // LatestEventScanner

// stream the latest 100 events in the `from..=to` block range
let mut event_stream = scanner.stream().await?; 
```

This is because internally, the `LatestEventScanner` type would have default values for `from_block` and `to_block` as `Earliest` and `Latest` respectively. This block range is implicit anyway in the `Fetch Latest X Events` functionality, so by making these explicit, we allow for an additional use case.

---

Pseudo code for the implementation for reference:

```rust
// Base type - everything starts here
struct LiveEventScanner {
    url: String,
    confirmations: Option<u64>,
    filters: EventFilters,
}

impl LiveEventScanner {
    fn new(url: impl Into<String>) -> Self {
        // Self { url: url.into(), confirmations: None, filters: EventFilters::default() }
    }

    fn with_block_confirmations(mut self, count: u64) -> Self {
        // self.confirmations = Some(count); self
    }

    fn with_contract_address(mut self, address: Address) -> Self {
        // self.filters.address = Some(address); self
    }

    fn with_event(mut self, signature: B256) -> Self {
        // self.filters.events.push(signature); self
    }

    fn from_block(self, block: BlockNumberOrTag) -> SyncEventScanner {
        // SyncEventScanner { url: self.url, from_block: block, ... }
    }

    fn latest(self, count: u64) -> LatestEventScanner {
        // LatestEventScanner { url: self.url, count, ... }
    }

    async fn stream(self) -> Result<EventStream> {
        // subscribe to new events from now
    }
}

// Calling .from_block() transitions to sync mode
struct SyncEventScanner {
    url: String,
    from_block: BlockNumberOrTag,
    confirmations: Option<u64>,
    filters: EventFilters,
}

impl SyncEventScanner {
    fn with_block_confirmations(mut self, count: u64) -> Self {
        // self.confirmations = Some(count); self
    }

    fn with_contract_address(mut self, address: Address) -> Self {
        // self.filters.address = Some(address); self
    }

    fn with_event(mut self, signature: B256) -> Self {
        // self.filters.events.push(signature); self
    }

    fn to_block(self, block: BlockNumberOrTag) -> HistoricEventScanner {
        // HistoricEventScanner { url: self.url, from_block: self.from_block, to_block: block, ... }
    }

    async fn stream(self) -> Result<EventStream> {
        // sync from from_block, then subscribe to new events
    }
}

// Calling .to_block() on SyncEventScanner transitions to historic mode
struct HistoricEventScanner {
    url: String,
    from_block: BlockNumberOrTag,
    to_block: BlockNumberOrTag,
    // No confirmations field - not applicable for historic
    filters: EventFilters,
}

impl HistoricEventScanner {
    fn with_contract_address(mut self, address: Address) -> Self {
        // self.filters.address = Some(address); self
    }

    fn with_event(mut self, signature: B256) -> Self {
        // self.filters.events.push(signature); self
    }

    async fn stream(self) -> Result<EventStream> {
        // query events in block range from_block..=to_block
    }
}

// Calling .latest() on LiveEventScanner transitions to latest mode
struct LatestEventScanner {
    url: String,
    count: u64,
    from_block: BlockNumberOrTag, // defaults to Earliest
    to_block: BlockNumberOrTag,   // defaults to Latest
    filters: EventFilters,
}

impl LatestEventScanner {
    fn with_contract_address(mut self, address: Address) -> Self {
        // self.filters.address = Some(address); self
    }

    fn with_event(mut self, signature: B256) -> Self {
        // self.filters.events.push(signature); self
    }

    fn from_block(mut self, block: BlockNumberOrTag) -> Self {
        // self.from_block = block; self
    }

    fn to_block(mut self, block: BlockNumberOrTag) -> Self {
        // self.to_block = block; self
    }

    fn then_live(self) -> SyncLatestEventScanner {
        // SyncLatestEventScanner { url: self.url, count: self.count, ... }
    }

    async fn stream(self) -> Result<EventStream> {
        // query latest count events in block range from_block..=to_block
    }
}

// Calling .then_live() on LatestEventScanner transitions to sync-latest mode
struct SyncLatestEventScanner {
    url: String,
    count: u64,
    confirmations: Option<u64>,
    filters: EventFilters,
}

impl SyncLatestEventScanner {
    fn with_block_confirmations(mut self, count: u64) -> Self {
        // self.confirmations = Some(count); self
    }

    fn with_contract_address(mut self, address: Address) -> Self {
        // self.filters.address = Some(address); self
    }

    fn with_event(mut self, signature: B256) -> Self {
        // self.filters.events.push(signature); self
    }

    async fn stream(self) -> Result<EventStream> {
        // query latest count events, then subscribe to new events
    }
}

// Transitions:
// LiveEventScanner --from_block()--> SyncEventScanner
// SyncEventScanner --to_block()--> HistoricEventScanner
// LiveEventScanner --latest()--> LatestEventScanner
// LatestEventScanner --then_live()--> SyncLatestEventScanner
// LatestEventScanner --from_block()/to_block()--> LatestEventScanner (stays same type)
```

Related #104 https://github.com/OpenZeppelin/Event-Scanner/issues/113 https://github.com/OpenZeppelin/Event-Scanner/issues/10

Makes https://github.com/OpenZeppelin/Event-Scanner/issues/25 obsolete.

LeoPatOZ:
Resolves #105
Resolves #107
Resolves #104
Resolves #111
Resolves #49
Resolves #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@
 Event Scanner is a Rust library for streaming EVM-based smart contract events. It is built on top of the [`alloy`](https://github.com/alloy-rs/alloy) ecosystem and focuses on in-memory scanning without a backing database. Applications provide event filters; the scanner takes care of fetching historical ranges, bridging into live streaming mode, all whilst delivering the events as streams of data.
 
 ---
- 
 
 ## Table of Contents
 
@@ -76,8 +75,8 @@ async fn run_scanner(
         .connect_ws::<Ethereum>(ws_url).await?;
 
     let filter = EventFilter::new()
-        .with_contract_address(contract)
-        .with_event(MyContract::SomeEvent::SIGNATURE);
+        .contract_address(contract)
+        .event(MyContract::SomeEvent::SIGNATURE);
 
     let mut stream = scanner.subscribe(filter);
 
@@ -118,6 +117,7 @@ let scanner = EventScanner::sync()
 
 // Latest mode (recent blocks only)
 let scanner = EventScanner::latest()
+    .count(100)
     .block_read_limit(500)
     .connect_ws::<Ethereum>(ws_url).await?;
 ```
@@ -142,23 +142,23 @@ Create an `EventFilter` for each event stream you wish to process. The filter sp
 ```rust
 // Track a SPECIFIC event from a SPECIFIC contract
 let specific_filter = EventFilter::new()
-    .with_contract_address(*counter_contract.address())
-    .with_event(Counter::CountIncreased::SIGNATURE);
+    .contract_address(*counter_contract.address())
+    .event(Counter::CountIncreased::SIGNATURE);
 
 // Track a multiple events from a SPECIFIC contract
 let specific_filter = EventFilter::new()
-    .with_contract_address(*counter_contract.address())
-    .with_event(Counter::CountIncreased::SIGNATURE)
-    .with_event(Counter::CountDecreased::SIGNATURE);
+    .contract_address(*counter_contract.address())
+    .event(Counter::CountIncreased::SIGNATURE)
+    .event(Counter::CountDecreased::SIGNATURE);
 
 // Track a SPECIFIC event from a ALL contracts
 let specific_filter = EventFilter::new()
-    .with_event(Counter::CountIncreased::SIGNATURE);
+    .event(Counter::CountIncreased::SIGNATURE);
 
 // Track ALL events from a SPECIFIC contracts
 let all_contract_events_filter = EventFilter::new()
-    .with_contract_address(*counter_contract.address())
-    .with_contract_address(*other_counter_contract.address());
+    .contract_address(*counter_contract.address())
+    .contract_address(*other_counter_contract.address());
 
 // Track ALL events from ALL contracts in the block range
 let all_events_filter = EventFilter::new();
@@ -185,7 +185,7 @@ See integration tests under `tests/live_mode`, `tests/historic_mode`, and `tests
 
 ### Scanning Latest Events
 
-`scan_latest` collects the most recent matching events for each registered stream.
+Scanner mode that collects a specified number of the most recent matching events for each registered stream.
 
 - It does not enter live mode; it scans a block range and then returns.
 - Each registered stream receives at most `count` logs in a single message, chronologically ordered.
@@ -193,24 +193,23 @@ See integration tests under `tests/live_mode`, `tests/historic_mode`, and `tests
 Basic usage:
 
 ```rust
-use alloy::{eips::BlockNumberOrTag, network::Ethereum};
-use event_scanner::{EventFilter, event_scanner::{EventScanner, EventScannerMessage}};
+use alloy::{network::Ethereum, primitives::Address, transports::http::reqwest::Url};
+use event_scanner::{EventFilter, EventScanner, Message};
 use tokio_stream::StreamExt;
 
-async fn latest_example(ws_url: alloy::transports::http::reqwest::Url, addr: alloy::primitives::Address) -> eyre::Result<()> {
-    let mut client = EventScanner::new().connect_ws::<Ethereum>(ws_url).await?;
+async fn latest_events(ws_url: Url, addr: Address) -> anyhow::Result<()> {
+    let mut scanner = EventScanner::latest().count(10).connect_ws::<Ethereum>(ws_url).await?;
 
-    let filter = EventFilter::new().with_contract_address(addr);
-    let mut stream = client.subscribe(filter);
+    let filter = EventFilter::new().contract_address(addr);
+
+    let mut stream = scanner.subscribe(filter);
 
     // Collect the latest 10 events across Earliest..=Latest
-    client.scan_latest(10).await?;
+    scanner.start().await?;
 
     // Expect a single message with up to 10 logs, then the stream ends
-    while let Some(msg) = stream.next().await {
-        if let EventScannerMessage::Data(logs) = msg {
-            println!("Latest logs: {}", logs.len());
-        }
+    while let Some(Message::Data(logs)) = stream.next().await {
+        println!("Latest logs: {}", logs.len());
     }
 
     Ok(())
@@ -221,8 +220,11 @@ Restricting to a specific block range:
 
 ```rust
 // Collect the latest 5 events between blocks [1_000_000, 1_100_000]
-client
-    .scan_latest_in_range(5, BlockNumberOrTag::Number(1_000_000), BlockNumberOrTag::Number(1_100_000))
+let mut scanner = EventScanner::latest()
+    .count(5)
+    .from_block(1_000_000)
+    .to_block(1_100_000)
+    .connect_ws::<Ethereum>(ws_url).await?;
     .await?;
 ```
 
@@ -230,21 +232,21 @@ The scanner periodically checks the tip to detect reorgs. On reorg, the scanner 
 
 Notes:
 
-- Ensure you create streams via `subscribe()` before calling `scan_latest*` so listeners are registered.
+- Ensure you create streams via `subscribe()` before calling `start` so listeners are registered.
 <!-- TODO: uncomment once implemented - The function returns after delivering the messages; to continuously stream new blocks, use `scan_latest_then_live`. -->
 
 ---
 
 ## Examples
 
-- `examples/simple_counter` – minimal live-mode scanner using `EventScanner::live()`
+- `examples/live_scanning` – minimal live-mode scanner using `EventScanner::live()`
 - `examples/historical_scanning` – demonstrates replaying historical data using `EventScanner::historic()`
 - `examples/latest_events_scanning` – demonstrates scanning the latest events using `EventScanner::latest()`
 
 Run an example with:
 
 ```bash
-RUST_LOG=info cargo run -p simple_counter
+RUST_LOG=info cargo run -p live_scanning    
 # or
 RUST_LOG=info cargo run -p historical_scanning
 ```
--- a/tests/latest_events/basic.rs
+++ b/tests/latest_events/basic.rs
@@ -38,7 +38,7 @@ macro_rules! decrease {
 }
 
 #[tokio::test]
-async fn scan_latest_exact_count_returns_last_events_in_order() -> anyhow::Result<()> {
+async fn latest_scanner_exact_count_returns_last_events_in_order() -> anyhow::Result<()> {
     let count = 5;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let contract = setup.contract;
@@ -68,7 +68,7 @@ async fn scan_latest_exact_count_returns_last_events_in_order() -> anyhow::Resul
 }
 
 #[tokio::test]
-async fn scan_latest_fewer_available_than_count_returns_all() -> anyhow::Result<()> {
+async fn latest_scanner_fewer_available_than_count_returns_all() -> anyhow::Result<()> {
     let count = 5;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let contract = setup.contract;
@@ -90,7 +90,7 @@ async fn scan_latest_fewer_available_than_count_returns_all() -> anyhow::Result<
 }
 
 #[tokio::test]
-async fn scan_latest_no_events_returns_empty() -> anyhow::Result<()> {
+async fn latest_scanner_no_events_returns_empty() -> anyhow::Result<()> {
     let count = 5;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let scanner = setup.scanner;
@@ -107,7 +107,7 @@ async fn scan_latest_no_events_returns_empty() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
-async fn scan_latest_respects_range_subset() -> anyhow::Result<()> {
+async fn latest_scanner_respects_range_subset() -> anyhow::Result<()> {
     let (anvil, provider, contract, default_filter) = setup_common(None, None).await?;
     // Mine 6 events, one per tx (auto-mined), then manually mint 2 empty blocks to widen range
     _ = increase!(contract);
@@ -144,7 +144,8 @@ async fn scan_latest_respects_range_subset() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
-async fn scan_latest_multiple_listeners_to_same_event_receive_same_results() -> anyhow::Result<()> {
+async fn latest_scanner_multiple_listeners_to_same_event_receive_same_results() -> anyhow::Result<()>
+{
     let count = 5;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let contract = setup.contract;
@@ -181,7 +182,7 @@ async fn scan_latest_multiple_listeners_to_same_event_receive_same_results() -> 
 }
 
 #[tokio::test]
-async fn scan_latest_different_filters_receive_different_results() -> anyhow::Result<()> {
+async fn latest_scanner_different_filters_receive_different_results() -> anyhow::Result<()> {
     let count = 3;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let contract = setup.contract;
@@ -228,7 +229,7 @@ async fn scan_latest_different_filters_receive_different_results() -> anyhow::Re
 }
 
 #[tokio::test]
-async fn scan_latest_mixed_events_and_filters_return_correct_streams() -> anyhow::Result<()> {
+async fn latest_scanner_mixed_events_and_filters_return_correct_streams() -> anyhow::Result<()> {
     let count = 2;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
     let contract = setup.contract;
@@ -271,7 +272,7 @@ async fn scan_latest_mixed_events_and_filters_return_correct_streams() -> anyhow
 }
 
 #[tokio::test]
-async fn scan_latest_cross_contract_filtering() -> anyhow::Result<()> {
+async fn latest_scanner_cross_contract_filtering() -> anyhow::Result<()> {
     // Manual setup to deploy two contracts
     let count = 5;
     let setup = setup_latest_scanner(None, None, count, None, None).await?;
@@ -305,7 +306,7 @@ async fn scan_latest_cross_contract_filtering() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
-async fn scan_latest_large_gaps_and_empty_ranges() -> anyhow::Result<()> {
+async fn latest_scanner_large_gaps_and_empty_ranges() -> anyhow::Result<()> {
     // Manual setup to mine empty blocks
     let (anvil, provider, contract, default_filter) = setup_common(None, None).await?;
 
@@ -340,7 +341,7 @@ async fn scan_latest_large_gaps_and_empty_ranges() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
-async fn scan_latest_boundary_range_single_block() -> anyhow::Result<()> {
+async fn latest_scanner_boundary_range_single_block() -> anyhow::Result<()> {
     let (anvil, provider, contract, default_filter) = setup_common(None, None).await?;
 
     _ = increase!(contract);


README.md
```diff
@@ -76,9 +76,8 @@ async fn run_scanner(
 
     let mut stream = client.create_event_stream(filter);
 
-    tokio::spawn(async move {
-        client.start_scanner(BlockNumberOrTag::Earliest, Some(BlockNumberOrTag::Latest)).await
-    });
+    // Live mode with default confirmations
+    tokio::spawn(async move { client.stream_live(Option::None).await });
```
0xNeshi, line 80: A couple of comments/thoughts:

1. I like that you took the opportunity to separate modes into different commands :heavy_check_mark: 

--- 

2. Ah, I find this API for passing configs very dev-UN-friendly, especially when multiple configs are needed (`client.stream_from(start, reads_per_epoch: Option<usize>, block_confirmations: Option<u64>)`)

Did you consider any other API approaches, e.g. builder pattern that we already use?
I understand that a builder pattern for a single config makes little sense, but it is more expressive in general (what does `stream_from(Option::None, Option::None, Option::Some(5))` represent anyway) and would allow us to add new configs if/when necessary later :thinking: 

---

3. I've been thinking about the following API redesign for a couple of days, interested to hear your thoughts. It might necessitate a larger API overhaul, but it would be much nicer to use if the client only needed to do:
```rust
let filter = EventFilter::new()
    .with_contract_address(contract)
    .with_event(MyContract::SomeEvent::SIGNATURE)
    .with_block_confirmations(5);

let mut stream = client.stream_live(filter).await;
```
Benefits:
- intuitive
- clean
- no need for manually spawning a task you'll rarely have to handle

Regarding hiding `tokio::spawn`, it must be spawned every time, so makes sense that we do it internally instead of forcing the boilerplate on the client)

This would only work if:
1. We repurpose `EventFilter` to a general event stream config struct (not just event-specific); maybe a different struct name would be necessary.
2. Any errors during stream init could be streamed back from within `tokio::spawn(...)` code

README.md
```diff
@@ -94,13 +93,16 @@ async fn run_scanner(
 
 ### Building a Scanner
 
-`EventScanner` supports:
+- No builders. Pass parameters to each mode as options.
```
0xNeshi, line 96: Updated docs immediately, nice :+1:

src/event_scanner.rs
```diff
@@ -57,40 +57,19 @@ impl EventScanner {
...
-    pub async fn connect_ws<N: Network>(self, ws_url: Url) -> Result<Client<N>, EventScannerError> {
+    pub async fn connect_ws<N: Network>(
+        self,
+        ws_url: Url,
+    ) -> Result<EventScannerClient<N>, EventScannerError> {
         let block_range_scanner = self.block_range_scanner.connect_ws(ws_url).await?;
         let event_scanner =
             ConnectedEventScanner { block_range_scanner, event_listeners: Vec::default() };
-        Ok(Client { event_scanner })
+        Ok(EventScannerClient { event_scanner })
```
0xNeshi, line 72: Why the rename?

In the local context, `Client` should be clear and succinct enough. If a dev importing this needs a more expressive name, they could alias it

src/event_scanner.rs
```diff
@@ -130,27 +109,73 @@ pub struct ConnectedEventScanner<N: Network> {
...
+    ) -> Result<(), EventScannerError> {
+        let client = self.block_range_scanner.run()?;
+        let mut stream = client.stream_live(block_confirmations).await?;
+
+        let (range_tx, _) = broadcast::channel::<BlockRangeMessage>(1024);
+        self.spawn_log_consumers(&range_tx);
+
+        while let Some(message) = stream.next().await {
+            if let Err(err) = range_tx.send(message) {
+                error!(error = %err, "failed sending message to broadcast channel");
```
0xNeshi, line 129: 1. Shouldn't this break the while loop since the channel is closed?
2. Add more detail to the error message

src/event_scanner.rs
```diff
@@ -130,27 +109,73 @@ pub struct ConnectedEventScanner<N: Network> {
...
-        };
+        let mut stream =
+            client.stream_historical(start_height, end_height, reads_per_epoch).await?;
 
         let (range_tx, _) = broadcast::channel::<BlockRangeMessage>(1024);
+        self.spawn_log_consumers(&range_tx);
+
+        while let Some(message) = stream.next().await {
+            if let Err(err) = range_tx.send(message) {
+                error!(error = %err, "failed sending message to broadcast channel");
```
0xNeshi, line 156: Same as before

src/event_scanner.rs
```diff
@@ -130,27 +109,73 @@ pub struct ConnectedEventScanner<N: Network> {
...
+
+        Ok(())
+    }
 
+    /// Implementation of stream from
+    ///
+    /// # Errors
+    ///
+    /// * `EventScannerMessage::ServiceShutdown` - if the service is already shutting down.
+    pub async fn stream_from(
```
0xNeshi, line 168: Update the while loop in `stream_from` to break the loop on error sending to the broadcast channel

src/event_scanner.rs
```diff
@@ -57,40 +57,19 @@ impl EventScanner {
...
-    #[must_use]
-    pub fn with_reorg_rewind_depth(mut self, reorg_rewind_depth: u64) -> Self {
-        self.block_range_scanner =
-            self.block_range_scanner.with_reorg_rewind_depth(reorg_rewind_depth);
-        self
-    }
-
-    /// Configures how many confirmations are required before processing a block (used for reorgs).
-    #[must_use]
-    pub fn with_block_confirmations(mut self, block_confirmations: u64) -> Self {
```
0xNeshi, line 78: Block confirmation should be a global config, as per #107

src/block_range_scanner.rs
```diff
@@ -190,22 +155,7 @@ impl BlockHashAndNumber {
...
-        dead_code,
-        reason = "Will be used in reorg mechanism: https://github.com/OpenZeppelin/Event-Scanner/issues/5"
-    )]
-    block_confirmations: u64,
-}
-
-pub struct BlockRangeScanner {
-    blocks_read_per_epoch: usize,
-    reorg_rewind_depth: u64,
-    block_confirmations: u64,
```
0xNeshi, line 207: Should be part of global config, as per #107

0xNeshi (commented):
First pass, interested in your thoughts on the API proposal below :eyes:

src/event_scanner.rs
LeoPatOZ: I think you meant blcok read per epoch. Block confirmations are still only relevant to live and sync mode (even what the issue states_

src/block_range_scanner.rs
LeoPatOZ: updated

src/event_scanner.rs
LeoPatOZ: okay changed it back to Client - i remember when trying to change things having this generic name was annoying so I always aliased it but we can do this later if really needed.

README.md
LeoPatOZ: 2. I agree the api is clunky - perhaps builder is more dev friendly. I did consider some streaming struct like:
```rust
let streaming_config = Config {
     block_confirmation: Some(5)
}
client.stream_live(streaming_config).await 
```
However each streaming mode would require a different config or you would pass in unneeded parameters (like start_block for live mode....

3. I like the API but similar to point 2. it could lead to having configs that dont make sense in certain modes like 
```rust
let filter = EventFilter::new()
    .with_contract_address(contract)
    .with_event(MyContract::SomeEvent::SIGNATURE)
    .with_block_confirmations(5); // this does nothing in historical mode
let mut stream = client.stream_historical(filter).await;
```
If we're fine with this maybe its nicer.

README.md
```diff
@@ -60,27 +60,29 @@ event-scanner = "0.3.0-alpha"
...
         .with_event(MyContract::SomeEvent::SIGNATURE);
 
-    let mut stream = client.create_event_stream(filter);
+    let mut stream = scanner.create_event_stream(filter);
 
-    tokio::spawn(async move {
-        client.start_scanner(BlockNumberOrTag::Earliest, Some(BlockNumberOrTag::Latest)).await
-    });
+    // Start the scanner
+    tokio::spawn(async move { scanner.stream().await });
```
0xNeshi: Move task spawning inside the `scanner.stream().await?` call. Rely on oneshot channels for scanner setup-related error propagation out of the task

examples/historical_scanning/main.rs
```diff
@@ -1,23 +1,19 @@
...
-contract Counter {
+    #[sol(rpc,
+bytecode="608080604052346015576101b0908161001a8239f35b5f80fdfe6080806040526004361015610012575f80fd5b5f3560e01c90816306661abd1461016157508063a87d942c14610145578063d732d955146100ad5763e8927fbc14610048575f80fd5b346100a9575f3660031901126100a9575f5460018101809111610095576020817f7ca2ca9527391044455246730762df008a6b47bbdb5d37a890ef78394535c040925f55604051908152a1005b634e487b7160e01b5f52601160045260245ffd5b5f80fd5b346100a9575f3660031901126100a9575f548015610100575f198101908111610095576020817f53a71f16f53e57416424d0d18ccbd98504d42a6f98fe47b09772d8f357c620ce925f55604051908152a1005b60405162461bcd60e51b815260206004820152601860248201527f436f756e742063616e6e6f74206265206e6567617469766500000000000000006044820152606490fd5b346100a9575f3660031901126100a95760205f54604051908152f35b346100a9575f3660031901126100a9576020905f548152f3fea2646970667358221220b846b706f79f5ae1fc4a4238319e723a092f47ce4051404186424739164ab02264736f6c634300081e0033"
```
0xNeshi: Add a comment above specifying which compiler version was used and on what platform, e.g.:
```
    // Built with Remix IDE; solc 0.8.24+commit.e11b9ed9
```

README.md
0xNeshi: Closing in favor of https://github.com/OpenZeppelin/Event-Scanner/pull/108#discussion_r2426664611

src/block_range_scanner.rs
```diff
@@ -93,18 +89,13 @@ use tracing::{debug, error, info, warn};
 pub const DEFAULT_BLOCKS_READ_PER_EPOCH: usize = 1000;
```
0xNeshi:
```suggestion
pub const DEFAULT_MAX_BLOCK_RANGE: usize = 1000;
```

src/block_range_scanner.rs
```diff
@@ -754,6 +734,7 @@ impl<N: Network> Service<N> {
         provider: P,
         sender: mpsc::Sender<BlockRangeMessage>,
         block_confirmations: u64,
+        max_read_per_epoch: usize,
```
0xNeshi: Can you try updating this to `u64` to avoid the conversions we often perform?

src/event_lib/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    transports::{RpcError, TransportErrorKind},
+};
+use thiserror::Error;
+
+use crate::{EventScannerMessage, block_range_scanner::BlockRangeScannerError};
+
+#[derive(Error, Debug, Clone)]
+pub enum EventScannerError {
+    #[error("Block range scanner error: {0}")]
+    BlockRangeScanner(#[from] BlockRangeScannerError),
```
0xNeshi: Is it possible to flatten this enum?

It's super annoying having to do:
```rust
match err {
    EventScannerError::BlockRangeScanner(e) => {
        match e {
            BlockRangeScannerError::ChannelError => { ... }
            ...
        }
    Provider(e) => { ... }
}
```
or even worse:
```rust
match err {
    EventScannerError::BlockRangeScanner(BlockRangeScannerError::ChannelError) => { ... }
    EventScannerError::BlockRangeScanner(BlockRangeScannerError::ServiceShutdown) => { ... }
    ...
    Provider(e) => { ... }
}
```

Would be much nicer to do:
```rust
match err {
    EventScannerError::ChannelError => { ... }
    EventScannerError::::ServiceShutdown => { ... }
    ...
    Provider(e) => { ... }
}
```

Especially because `BlockRangeScannerError` and `EventScannerError` share the same `RpcError(Arc<RpcError<TransportErrorKind>>)` and `Provider(Arc<RpcError<TransportErrorKind>>)` respectively. 
Related suggestion, maybe we should align the names? Whichever one you prefer, both are fine to me.

If there's no macro to do this flattening, I'd be in favor copy/pasting errors from `BlockRangeScannerError`. We'd then need to manually implement the `From` conversion, which would ensure that event scanner errors never get out of sync when `BlockRangeScannerError` enum is updated.

src/block_range_scanner.rs
```diff
@@ -293,32 +225,25 @@ impl BlockRangeScanner {
...
     ) -> TransportResult<ConnectedBlockRangeScanner<N>> {
-        Ok(ConnectedBlockRangeScanner {
-            provider,
-            config: Config {
-                blocks_read_per_epoch: self.blocks_read_per_epoch,
-                reorg_rewind_depth: self.max_reorg_depth,
-                block_confirmations: self.block_confirmations,
-            },
-        })
+        Ok(ConnectedBlockRangeScanner { provider, max_read_per_epoch: self.max_block_range })
```
0xNeshi: This can never error out, so it should return `ConnectedBlockRangeScanner` directly.

src/event_lib/modes/historic.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the connection fails
+    pub async fn connect_ws<N: Network>(
+        self,
+        ws_url: Url,
+    ) -> TransportResult<HistoricEventScanner<N>> {
+        let HistoricScannerConfig { base, from_block, to_block } = self;
+        let brs = base.block_range_scanner.connect_ws::<N>(ws_url).await?;
+        let config = HistoricScannerConfig { base, from_block, to_block };
```
0xNeshi: Why the unpack/repack?

src/event_lib/modes/historic.rs
```diff
@@ -0,0 +1,165 @@
...
+        let config =
+            HistoricScannerConfig::new().from_block(100u64).to_block(200u64).max_block_range(50);
+
+        assert!(matches!(config.from_block, BlockNumberOrTag::Number(100)));
+        assert!(matches!(config.to_block, BlockNumberOrTag::Number(200)));
+        assert_eq!(config.base.block_range_scanner.max_block_range, 50);
+    }
+
+    #[test]
+    fn test_historic_scanner_builder_pattern_chaining() {
```
0xNeshi: What's different in this test from the previous one (`test_historic_scanner_builder_pattern`)?

src/event_lib/modes/latest.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub fn to_block(mut self, block: impl Into<BlockNumberOrTag>) -> Self {
+        self.to_block = block.into();
+        self
+    }
+
+    #[must_use]
+    pub fn then_live(mut self) -> Self {
+        self.switch_to_live = true;
+        self
+    }
```
0xNeshi: This pattern has the same problems as the original proposal I submitted when we started talking about upgrading the API: it is possible to call `then_live` on a scanner builder that already has `to_block` set, which would:
1. Discard the `to_block` value, as it does nothing in `latest then live` mode.
2. Confuse devs.

We'd either need a separate `latest_then_live` scanner, or we'd need to upgrade this scanner to _not expose_ this function if `to_block` is set. To achieve this, you could utilize the so-called "type-state pattern", e.g.:
```rust
pub struct NeedsLimit;
pub struct Complete;

pub struct CounterBuilder<T> {
    limit: u64,
    has: PhantomData<T>,
}

impl CounterBuilder<NeedsLimit> {
    pub fn new() -> Self {
        Self { limit: 0, has: PhantomData }
    }

    pub fn limit(self, limit: u64) -> Result<CounterBuilder<Complete>, Error> {
        if limit == 0 {
            return Err(Error::ZeroLimit);
        }
        Ok(CounterBuilder {
            limit,
            has: PhantomData,
        })
    }
}
```

src/event_lib/modes/sync.rs
```diff
@@ -0,0 +1,177 @@
...
+};
+
+use super::{BaseConfig, BaseConfigBuilder};
+
+pub struct SyncScannerConfig {
+    base: BaseConfig,
+    // Defaults to Earliest
+    from_block: BlockNumberOrTag,
+    // Defaults to 0
+    block_confirmations: u64,
```
0xNeshi: [Defaults](https://github.com/OpenZeppelin/Event-Scanner/pull/108/files#diff-6224cab175b6335330fe0c13ef7d1ee2ea0d971afe5ce6f8b1448f894f676b30R45) to `DEFAULT_BLOCK_CONFIRMATIONS`

src/event_scanner/modes/live.rs
```diff
@@ -0,0 +1,143 @@
...
+impl LiveScannerConfig {
+    pub(super) fn new() -> Self {
+        Self { base: BaseConfig::new(), block_confirmations: DEFAULT_BLOCK_CONFIRMATIONS }
+    }
+
+    #[must_use]
+    pub fn block_confirmations(mut self, count: u64) -> Self {
+        self.block_confirmations = count;
+        self
+    }
```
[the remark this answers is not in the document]
0xNeshi: Or don't, we can do it as part of #50

examples/sync_scanning/main.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+
+        if historical_processed && live_processed {
+            info!("Both historical and live events processed successfully!");
+            break;
+        }
+    }
+
+    scanner_task.abort();
```
0xNeshi:
```suggestion
```
nit: will abort when dropped

examples/sync_scanning/main.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let mut stream = scanner.create_event_stream(increase_filter);
+
+    info!("Starting sync scanner...");
+    let scanner_task = tokio::spawn(async move {
+        scanner.start().await.expect("failed to start scanner");
+    });
+
+    // Wait a bit for historical events to be processed
+    sleep(Duration::from_secs(2)).await;
```
0xNeshi: nit: the scanner should process events even without waiting

src/event_lib/scanner.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// # Reorg behavior
+    ///
+    /// * Emits [`ScannerStatus::ReorgDetected`] and adjusts the next block range using
+    ///   `block_confirmations` to re-emit the confirmed portion.
+    ///
+    /// # Errors
+    ///
+    /// * `EventScannerMessage::ServiceShutdown` - if the service is already shutting down.
+    pub async fn stream_live(&self, block_confirmations: u64) -> Result<(), EventScannerError> {
```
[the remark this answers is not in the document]
LeoPatOZ: actually should have done that from the beginning. Removed the client now we just handle everything in scanners themselves

src/event_scanner/modes/historic.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.to_block = block.into();
+        self
+    }
+
+    /// Connects to the provider via WebSocket
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the connection fails
+    pub async fn connect_ws<N: Network>(
```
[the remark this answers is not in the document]
LeoPatOZ: yeah left it for now

src/event_lib/modes/sync.rs
```diff
@@ -0,0 +1,177 @@
...
+        filter::EventFilter,
+        scanner::{EventScannerMessage, EventScannerService},
+    },
+};
+
+use super::{BaseConfig, BaseConfigBuilder};
+
+pub struct SyncScannerConfig {
+    base: BaseConfig,
+    // Defaults to Earliest
```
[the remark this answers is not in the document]
0xNeshi: Remove in all of them.

Since this is private code (lib users shouldn't be able to access these structs directly), the comments make little sense, as we (the lib devs) know the defaults or can easily find them in the `new` call.

src/block_range_scanner.rs
```diff
@@ -64,8 +60,7 @@
 //! }
 //! ```
 
-use std::{cmp::Ordering, ops::RangeInclusive, sync::Arc};
-
+use std::{cmp::Ordering, ops::RangeInclusive, result::Result::Ok, sync::Arc};
```
0xNeshi:
```suggestion
use std::{cmp::Ordering, ops::RangeInclusive, sync::Arc};
```

src/block_range_scanner.rs
```diff
@@ -425,22 +380,25 @@ impl<N: Network> Service<N> {
...
             Command::Unsubscribe { response } => {
                 self.handle_unsubscribe();
-                let _ = response.send(Ok(()));
+                let _ = response.send(Ok::<(), BlockRangeScannerError>(()));
             }
             Command::Shutdown { response } => {
                 self.shutdown = true;
                 self.handle_unsubscribe();
-                let _ = response.send(Ok(()));
+                let _ = response.send(Ok::<(), BlockRangeScannerError>(()));
```
0xNeshi: Why the expansion?

src/event_scanner/message.rs
```diff
@@ -0,0 +1,39 @@
+use alloy::{rpc::types::Log, sol_types::SolEvent};
+
+use crate::{EventScannerError, ScannerMessage};
+
+pub type EventScannerMessage = ScannerMessage<Vec<Log>, EventScannerError>;
```
0xNeshi: Similar as for errors, let's shorten this to just `Message`. It's obvious that this is an event scanner message; it is in the `event_scanner` module after all

src/event_scanner/modes/historic.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    #[test]
+    fn test_historic_scanner_builder_pattern_random_order_chaining() {
+        let config =
+            HistoricScannerBuilder::new().to_block(200).max_block_range(50).from_block(100);
+
+        assert!(matches!(config.from_block, BlockNumberOrTag::Number(100)));
+        assert!(matches!(config.to_block, BlockNumberOrTag::Number(200)));
+        assert_eq!(config.block_range_scanner.max_block_range, 50);
+    }
```
0xNeshi: nit: I feel like this may be a redundant test after all. It's not like we anticipate build order to affect the final result :thinking:

src/event_scanner/modes/live.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+pub struct LiveEventScanner<N: Network> {
+    config: LiveScannerBuilder,
+    block_range_scanner: ConnectedBlockRangeScanner<N>,
+    listeners: Vec<EventListener>,
+}
+
+impl LiveScannerBuilder {
+    pub(super) fn new() -> Self {
```
0xNeshi: Shouldn't this be `pub(crate)`? Do we want to allow lib users to manually instantiate builders?

src/event_scanner/modes/historic.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+pub struct HistoricEventScanner<N: Network> {
+    config: HistoricScannerBuilder,
+    block_range_scanner: ConnectedBlockRangeScanner<N>,
+    listeners: Vec<EventListener>,
+}
+
+impl HistoricScannerBuilder {
+    pub(super) fn new() -> Self {
```
0xNeshi:
```suggestion
    #[must_use]
    pub(super) fn new() -> Self {
```
applies to other builders' `new`

0xNeshi (commented):
Looks very good, left an additional batch of change requests, I think we can merge afterwards

src/event_scanner/modes/historic.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Returns an error if the connection fails
+    #[must_use]
+    pub fn connect<N: Network>(self, provider: RootProvider<N>) -> HistoricEventScanner<N> {
+        let block_range_scanner = self.block_range_scanner.connect::<N>(provider);
+        HistoricEventScanner { config: self, block_range_scanner, listeners: Vec::new() }
+    }
+}
+
+impl<N: Network> HistoricEventScanner<N> {
+    pub fn create_event_stream(
```
0xNeshi:
```suggestion
    pub fn subscribe(
```
this might be clear enough, wdyt?

Issue #118, opened 2025-10-07
fix: Ensure Channel Senders Are Appropriately Dropped after Streaming is Done


0xNeshi:
E.g. in sync mode, historic mode sender should be dropped after the catch-up phase is done and `ChainTipReached` is streamed.

Issue #10, opened 2025-09-11
Add 'Get Latest X Events' Query Method


LeoPatOZ:
### Description:
Implement method to fetch the most recent X events without full historical scanning.

Implementation:
- Query method to get last N events
- Efficient backward block scanning
- Optional filter by event type

Potential API:
```rust
scanner.get_latest_events(count: usize, filter: Option<EventFilter>)
```

Maybe this could split the task in multiple batches (scan the latest 500 blocks in 50 block batches) tbd..

Use case: Taiko needs this for Shasta because they're need the past 2 proposals for some of their logic.

0xNeshi:
- the events should be streamed in the order that they were emitted
- streams without regard to reorgs

PR #119, opened 2025-10-07
fix: Fix block range

0xNeshi:
Towards #118

Took the opportunity to fix this while working on #10


LeoPatOZ (approved):
Happy to merge once the test is fixed

Issue #112, opened 2025-10-06
feat: External RPC Calls Should be Wrapped in Retry + Backoff Mechanisms


0xNeshi:
If the retry mechanism fails X number of times, bubble up the error through the stream so the client can determine what to do.

Each failed call/action should be logged.

Issue #66, opened 2025-09-25
test: Add Robustness Tests


0xNeshi:
Some of the necessary test cases:
- [ ] Slow RPC response #135 
- [ ] Failed RPC response #135 
- [ ] Slow event processing, causing channel to be filled

PR #135, opened 2025-10-20
Retry logic

LeoPatOZ:
Resolves https://github.com/OpenZeppelin/Event-Scanner/issues/112
Towards #66

1. Do we want to use any other features of backoff [here](https://github.com/OpenZeppelin/Event-Scanner/blob/retry-logic/src/safe_provider.rs#L201)?
2. Is this sufficient logging?

### DIFF
--- a/src/block_range_scanner.rs
+++ b/src/block_range_scanner.rs
@@ -424,23 +424,14 @@ impl<N: Network> Service<N> {
         let get_start_block = async || -> Result<BlockNumber, ScannerError> {
             let block = match start_height {
                 BlockNumberOrTag::Number(num) => num,
-                block_tag => provider
-                    .get_block_by_number(block_tag)
-                    .await?
-                    .ok_or_else(|| ScannerError::BlockNotFound(block_tag))?
-                    .header()
-                    .number(),
+                block_tag => provider.get_block_by_number(block_tag).await?.header().number(),
             };
             Ok(block)
         };
 
         let get_latest_block = async || -> Result<BlockNumber, ScannerError> {
-            let block = provider
-                .get_block_by_number(BlockNumberOrTag::Latest)
-                .await?
-                .ok_or_else(|| ScannerError::BlockNotFound(BlockNumberOrTag::Latest))?
-                .header()
-                .number();
+            let block =
+                provider.get_block_by_number(BlockNumberOrTag::Latest).await?.header().number();
             Ok(block)
         };
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1085,6 +1085,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "backon"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cffb0e931875b666fc4fcb20fee52e9bbd1ef836fd9e9e04ec21555f9f85f7ef"
+dependencies = [
+ "fastrand",
+ "gloo-timers",
+ "tokio",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.75"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1667,6 +1678,7 @@ dependencies = [
  "alloy-node-bindings",
  "anyhow",
  "async-trait",
+ "backon",
  "chrono",
  "serde",
  "serde_json",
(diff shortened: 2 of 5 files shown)


src/block_range_scanner.rs
```diff
@@ -269,6 +276,24 @@ impl BlockRangeScanner {
         self
     }
 
+    #[must_use]
+    pub fn with_max_timeout(mut self, rpc_timeout: Duration) -> Self {
```
0xNeshi: remove `with_` prefixes

src/safe_provider.rs
```diff
@@ -0,0 +1,344 @@
...
+    #[must_use]
+    pub fn with_retry_interval(mut self, retry_interval: Duration) -> Self {
+        self.retry_interval = retry_interval;
+        self
+    }
+
+    #[must_use]
+    pub fn inner(&self) -> &RootProvider<N> {
+        &self.provider
+    }
```
0xNeshi: Since this wrapper will be perfect for abstracting away multi-provider support (see #6), this function will become redundant. Being event currently unused, remove it until it becomes actually necessary.

src/safe_provider.rs
```diff
@@ -0,0 +1,344 @@
...
+        &self,
+        number: BlockNumberOrTag,
+    ) -> Result<Option<N::BlockResponse>, RpcError<TransportErrorKind>> {
+        debug!("SafeProvider eth_getBlockByNumber called with number: {:?}", number);
+        let provider = self.provider.clone();
+        let result = self
+            .retry_with_total_timeout(|| async { provider.get_block_by_number(number).await })
+            .await;
+        if let Err(e) = &result {
+            error!("SafeProvider eth_getByBlockNumber failed: {}", e);
```
0xNeshi:
```suggestion
            error!(error = e, "SafeProvider eth_getByBlockNumber failed");
```
Also, maybe no need to mention the underlying `SafeProvider` type in logs? Just a notification that an operation failed should be sufficient

src/safe_provider.rs
```diff
@@ -0,0 +1,344 @@
...
+        let provider = create_test_provider(max_timeout, 10, Duration::from_millis(1));
+
+        let result = provider
+            .retry_with_total_timeout(move || async move {
+                sleep(max_timeout + Duration::from_millis(10)).await;
+                Ok(42)
+            })
+            .await;
+
+        assert!(result.is_err());
```
0xNeshi: verify that it matches the expected timeout error

src/safe_provider.rs
LeoPatOZ: You cant do the error = e because error needs to implement some like tracing trait thing 

Yeah i added it because i wasnt sure if it was obvious where the error came from if it was just like eth_getByBlockNumber failed..

Removed though perhaps its useless :)

src/safe_provider.rs
LeoPatOZ: same as above

src/safe_provider.rs
```diff
@@ -0,0 +1,344 @@
...
+//!         ClientBuilder::default().ws(WsConnect::new("wss://localhost:8000")).await?,
+//!     );
+//!     let safe_provider = SafeProvider::new(provider)
+//!         .with_max_timeout(Duration::from_secs(30))
+//!         .with_max_retries(5);
+//!
+//!     let block = safe_provider.get_block_by_number(12345.into()).await?;
+//!     Ok(())
+//! }
+//! ```
```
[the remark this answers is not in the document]
LeoPatOZ: Actually removed the doc completely - i dont think we should expose the safe provider to outside the crate, therefore it doesnt make sense

src/robust_provider.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// Fetch a block by hash with retry and timeout.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if RPC call fails repeatedly even
+    /// after exhausting retries or if the call times out.
+    pub async fn get_block_by_hash(
+        &self,
+        hash: alloy::primitives::BlockHash,
+    ) -> Result<Option<N::BlockResponse>, RobustProviderError> {
```
0xNeshi: Update this similarly to `get_block_by_number`

src/robust_provider.rs
```diff
@@ -0,0 +1,312 @@
...
+        };
+        let result = self.retry_with_total_timeout(operation).await;
+        if let Err(e) = &result {
+            error!(error = %e, "eth_getByBlockNumber failed");
+        }
+
+        match result? {
+            Some(block) => Ok(block),
+            None => Err(RobustProviderError::BlockNotFound(number)),
+        }
```
0xNeshi:
```suggestion
        result?.ok_or(Err(RobustProviderError::BlockNotFound(number)))
```

PR #137, opened 2025-10-23
build: rely on explicit test-utils feature setting

0xNeshi:
The Cargo.toml section doesn't really work, safer to rely simply on always passing the feature flag


LeoPatOZ:
Think we could add an alias to cargo like alias t or test to run `cargo nextest run --locked --all-targets --all-features`?

PR #140, opened 2025-10-27
Add Multiple providers

### LINKED ISSUE #6: Add Multiple Endpoints as Providers
### Description:
Add support for multiple RPC endpoints with automatic failover and load balancing to improve reliability and availability.

Implementation:

- Support multiple WebSocket/IPC endpoints
- Automatic failover on connection loss or errors
- Health checking for endpoints

Optional
- _Configurable failover strategies (round-robin, priority-based, latency-based) potentially_
- _Seamless reconnection without event loss_
- _Optional load balancing across healthy endpoints_

Potential API:
```rust 
ScannerBuilder::new_multi([
    Provider::primary("wss://main-node"),
    Provider::fallback("wss://backup-node"),
    Provider::fallback("https://http-backup"),
])
.failover_strategy(FailoverStrategy::Priority)
```

### LINKED ISSUE #142: feat: Add Support for HTTP RPC Providers
Historical streams (both block range and "latest events") can work with HTTP, as they're basically querying the chain data.

LeoPatOZ:
Resolves #6 
Resolves #142

### DIFF
--- a/src/robust_provider.rs
+++ b/src/robust_provider.rs
@@ -616,7 +616,8 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_subscribe_fails_if_primary_provider_lacks_pubsub() -> anyhow::Result<()> {
+    async fn test_subscribe_succeeds_if_primary_provider_lacks_pubsub_but_fallback_supports_it()
+    -> anyhow::Result<()> {
         let anvil = Anvil::new().try_spawn()?;
 
         let http_provider = ProviderBuilder::new().connect_http(anvil.endpoint_url());
@@ -630,17 +631,8 @@ mod tests {
             .build()
             .await?;
 
-        let err = robust.subscribe_blocks().await.unwrap_err();
-
-        match err {
-            Error::RpcError(e) => {
-                assert!(matches!(
-                    e.as_ref(),
-                    RpcError::Transport(TransportErrorKind::PubsubUnavailable)
-                ));
-            }
-            other => panic!("Expected PubsubUnavailable error type, got: {other:?}"),
-        }
+        let result = robust.subscribe_blocks().await;
+        assert!(result.is_ok());
 
         Ok(())
     }


src/event_scanner/scanner/mod.rs
```diff
@@ -365,9 +365,62 @@ impl EventScannerBuilder<SyncFromBlock> {
...
+    /// # let fallback_url = "ws://fallback:8545".parse()?;
+    /// let scanner = EventScannerBuilder::historic()
+    ///     .fallback_ws(fallback_url)
+    ///     .connect_ws::<Ethereum>(ws_url)
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[must_use]
+    pub fn fallback_ws(mut self, url: Url) -> Self {
```
0xNeshi: Differentiating between `main` and `fallback` providers makes little sense, because they're theoretically interchangeable **[1]**. It just introduces additional complexity, both implementation-wise and API-wise. 

It makes sense that the scanner would prioritize the first provider the user defined, but other than that, why not treat all providers the same?

Consider this API:
```rust
let ws_url = "ws://localhost:8545".parse()?;
let fallback_url = "ws://fallback:8545".parse()?;
let ipc_url = "ipc://whatever".parse?;

let scanner = EventScannerBuilder::<Ethereum>::historic()
    .provider_ws(ws_url)  // used as first, i.e. main provider
    .provider_ws(fallback_url) // implicitly a fallback provider
    .provider_ipc(ipc_url) // 2nd fallback
    .build()  // or even `.connect()`                            
    .await?;
```

---

**[1]**: This may change in the future, and we may add provider prioritization, implying the scanner would try to use the main provider as much as possible, and even if it temporarily switches to a fallback, it would occasionally try to reconnect to the main provider.

src/event_scanner/scanner/mod.rs
```diff
@@ -365,9 +365,62 @@ impl EventScannerBuilder<SyncFromBlock> {
...
+    /// # use alloy::network::Ethereum;
+    /// # use event_scanner::EventScannerBuilder;
+    /// #
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// # let ws_url = "ws://localhost:8545".parse()?;
+    /// # let fallback_url = "ws://fallback:8545".parse()?;
+    /// let scanner = EventScannerBuilder::historic()
+    ///     .fallback_ws(fallback_url)
+    ///     .connect_ws::<Ethereum>(ws_url)
+    ///     .await?;
```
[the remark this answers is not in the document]
LeoPatOZ: Actually was debating this but ended up choosing this way - which i see now is actually not great.

I think we should go with passing in the RobustProvider directly instead of muddling this api with both the scanner and provider methods.

tests/common/mod.rs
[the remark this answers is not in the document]
0xNeshi: Why?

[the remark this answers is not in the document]
LeoPatOZ: I provided a link but basically it's not recommended to have common.rs in the root test dir as cargo will see it as a test file and try to run it

src/robust_provider.rs
```diff
@@ -172,23 +199,76 @@ impl<N: Network> RobustProvider<N> {
...
+        Fut: Future<Output = Result<T, RpcError<TransportErrorKind>>>,
+    {
+        // Try primary provider first
+        let result = self.try_provider_with_timeout(&self.provider, &operation).await;
+
+        if let Ok(value) = result {
+            return Ok(value);
+        }
+
+        if result.is_err() && self.fallback_providers.is_empty() {
```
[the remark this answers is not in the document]
0xNeshi: You're right! My bad

src/event_scanner/scanner/mod.rs
```diff
@@ -467,16 +437,21 @@ mod tests {
...
         let _stream3 = scanner.subscribe(EventFilter::new());
         assert_eq!(scanner.listeners.len(), 3);
+
+        Ok(())
     }
 
-    #[test]
-    fn test_historic_event_stream_channel_capacity() {
+    #[tokio::test]
+    async fn test_historic_event_stream_channel_capacity() -> anyhow::Result<()> {
```
0xNeshi: nit: why the `anyhow::Result<()>`?

src/event_scanner/scanner/mod.rs
```diff
@@ -454,10 +423,11 @@ mod tests {
         assert_eq!(builder.config.block_confirmations, DEFAULT_BLOCK_CONFIRMATIONS);
     }
 
-    #[test]
-    fn test_historic_event_stream_listeners_vector_updates() {
+    #[tokio::test]
+    async fn test_historic_event_stream_listeners_vector_updates() -> anyhow::Result<()> {
```
0xNeshi: nit: why the `anyhow::Result<()>`?

src/robust_provider.rs
```diff
@@ -172,23 +213,76 @@ impl<N: Network> RobustProvider<N> {
...
```
0xNeshi:
```suggestion
        if self.fallback_providers.is_empty() {
```
nit: if the execution gets here, `result` [must be an error](https://github.com/OpenZeppelin/Event-Scanner/pull/140/files#diff-7c3abb68ead49d79c137d9cc6fde16d55052445640ed7eaaec0411546eea5276R233-R237)

src/robust_provider.rs
```diff
@@ -78,6 +87,21 @@ impl<N: Network> RobustProvider<N> {
...
+    #[must_use]
+    pub fn inner(&self) -> &RootProvider<N> {
+        &self.provider
+    }
+
+    /// Add a fallback provider to the list.
+    ///
+    /// Fallback providers are used when the primary provider times out.
+    #[must_use]
+    pub fn fallback_provider(mut self, provider: RootProvider<N>) -> Self {
```
0xNeshi:
```suggestion
    pub fn fallback(mut self, provider: RootProvider<N>) -> Self {
```
nit: clear enough + succinct

src/robust_provider.rs
```diff
@@ -172,23 +213,76 @@ impl<N: Network> RobustProvider<N> {
...
     {
         let retry_strategy = ExponentialBuilder::default()
             .with_max_times(self.max_retries)
             .with_min_delay(self.retry_interval);
 
-        match tokio::time::timeout(
+        match timeout(
             self.max_timeout,
-            operation.retry(retry_strategy).sleep(tokio::time::sleep),
+            (|| operation(provider.clone())).retry(retry_strategy).sleep(tokio::time::sleep),
```
0xNeshi: Set `backon` to notify on each retry, there's a dedicated function for that

src/robust_provider.rs
LeoPatOZ: true good catch

src/robust_provider.rs
```diff
@@ -172,23 +213,76 @@ impl<N: Network> RobustProvider<N> {
...
+        }
+
+        if result.is_err() && self.fallback_providers.is_empty() {
+            return result;
+        }
+
+        info!("Primary provider failed, trying fallback provider(s)");
+
+        // Try each fallback provider
+        for (idx, fallback_provider) in self.fallback_providers.iter().enumerate() {
```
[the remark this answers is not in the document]
LeoPatOZ: Yes true updated it now

src/robust_provider.rs
LeoPatOZ: https://github.com/OpenZeppelin/Event-Scanner/pull/140/commits/8b1079cc306be118b82a48fc16724ec5c39d5ae6 

Do you think this is fine ?

src/robust_provider.rs
```diff
@@ -172,23 +213,76 @@ impl<N: Network> RobustProvider<N> {
...
+    ///   providers succeed.
     /// - Propagates any [`RpcError<TransportErrorKind>`] from the underlying retries.
     async fn retry_with_total_timeout<T, F, Fut>(&self, operation: F) -> Result<T, Error>
     where
-        F: Fn() -> Fut,
+        F: Fn(RootProvider<N>) -> Fut,
+        Fut: Future<Output = Result<T, RpcError<TransportErrorKind>>>,
+    {
+        // Try primary provider first
+        let result = self.try_provider_with_timeout(&self.provider, &operation).await;
```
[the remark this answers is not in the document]
0xNeshi: solved in https://github.com/OpenZeppelin/Event-Scanner/pull/140/commits/655124ac842dddf5537cad08f0b8df455280510a

src/robust_provider.rs
```diff
@@ -10,71 +10,103 @@ use alloy::{
...
+        self
+    }
+
+    /// Get a reference to the primary provider (the first provider in the list)
+    ///
+    /// # Panics
+    ///
+    /// If there are no providers set (this should never happen)
+    #[must_use]
+    pub fn root(&self) -> &RootProvider<N> {
```
0xNeshi:
```suggestion
    /// Get a reference to the primary provider (the first provider in the list)
    ///
    /// # Panics
    ///
    /// If there are no providers set (this should never happen)
    #[must_use]
    pub fn primary(&self) -> &RootProvider<N> {
```
accurate name

src/robust_provider.rs
```diff
@@ -172,49 +226,129 @@ impl<N: Network> RobustProvider<N> {
...
+                    last_error = e;
+                }
+            }
+        }
+
+        // If all providers were skipped due to pubsub requirement
+        if skipped_count == self.providers.len() {
+            error!("All providers skipped - none support pubsub");
+            return Err(RpcError::Transport(TransportErrorKind::PubsubUnavailable).into());
+        }
```
0xNeshi: If all providers were skipped, the error returned should be from the primary provider (as it's the last valid pubsub-supporting provider)

src/robust_provider.rs
```diff
@@ -172,49 +226,129 @@ impl<N: Network> RobustProvider<N> {
...
+
+        // This loop starts at index 1 automatically
+        for (idx, provider) in providers.enumerate() {
+            let fallback_num = idx + 1;
+            if require_pubsub && !Self::supports_pubsub(provider) {
+                info!("Fallback provider {} doesn't support pubsub, skipping", fallback_num);
+                skipped_count += 1;
+                continue;
+            }
+            info!("Attempting fallback provider {}/{}", fallback_num, self.providers.len() - 1);
```
0xNeshi: nit: read `self.providers.len()` and store in a var outside the for loop

src/block_range_scanner.rs
```diff
@@ -161,62 +147,10 @@ impl BlockRangeScanner {
...
-    pub fn connect<N: Network>(self, provider: RootProvider<N>) -> ConnectedBlockRangeScanner<N> {
-        let robust_provider = RobustProvider::new(provider)
-            .max_timeout(self.max_timeout)
-            .max_retries(self.max_retries)
-            .retry_interval(self.retry_interval);
-        ConnectedBlockRangeScanner {
-            provider: robust_provider,
-            max_block_range: self.max_block_range,
-        }
+    pub fn connect<N: Network>(self, provider: RobustProvider<N>) -> ConnectedBlockRangeScanner<N> {
```
0xNeshi: Implement `From<Provider<N>> for RobustProvider<N>`. 
Then you make this accept `provider: impl Into<RobustProvider<N>>`, so that users can also pass alloy's `(Root)Provider<N>` here. This would also make our tests slightly less verbose to write.

Maybe it's safe to say that if a user is passing in `RootProvider`, that they don't wish for any advanced `RobustProvider` features, i.e. retry, fallbacks etc. In that case, `RobustProvider` should set only default timeouts.
Does that make sense to you?

0xNeshi: Related to previous, you can also implement `From<&str> for RobustProvider<N>` and utilize [alloy_provider::ProviderBuilder::connect](https://docs.rs/alloy-provider/latest/alloy_provider/struct.ProviderBuilder.html#method.connect) so that users can pass in any plain WebSocket or IPC connection strings or anything else that `alloy` supports for a provider. Very powerful.

src/event_scanner/scanner/mod.rs
```diff
@@ -389,42 +386,11 @@ impl<M> EventScannerBuilder<M> {
...
     /// Connects to an existing provider.
     ///
     /// Final builder method: consumes the builder and returns the built [`EventScanner`].
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if the connection fails
     #[must_use]
-    pub fn connect<N: Network>(self, provider: RootProvider<N>) -> EventScanner<M, N> {
+    pub fn connect<N: Network>(self, provider: RobustProvider<N>) -> EventScanner<M, N> {
```
0xNeshi: Same as in previous two change requests

src/event_scanner/scanner/mod.rs
```diff
@@ -478,10 +447,11 @@ mod tests {
         assert_eq!(builder.config.block_confirmations, DEFAULT_BLOCK_CONFIRMATIONS);
     }
 
-    #[test]
-    fn test_historic_event_stream_listeners_vector_updates() {
+    #[tokio::test]
+    async fn test_historic_event_stream_listeners_vector_updates() {
```
0xNeshi: Redundant, test is sync

src/robust_provider.rs
```diff
@@ -172,49 +226,129 @@ impl<N: Network> RobustProvider<N> {
...
+
+        // If all providers were skipped due to pubsub requirement
+        if skipped_count == self.providers.len() {
+            error!("All providers skipped - none support pubsub");
+            return Err(RpcError::Transport(TransportErrorKind::PubsubUnavailable).into());
+        }
+
+        // Return the last error encountered
+        error!("All providers failed or timed out");
+        Err(last_error)
```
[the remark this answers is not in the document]
LeoPatOZ: okay lmk if this is good!

src/robust_provider.rs
```diff
@@ -10,71 +10,103 @@ use alloy::{
...
 // RPC retry and timeout settings
 /// Default timeout used by `RobustProvider`
-pub const DEFAULT_MAX_TIMEOUT: Duration = Duration::from_secs(30);
+pub const DEFAULT_MAX_TIMEOUT: Duration = Duration::from_secs(60);
 /// Default maximum number of retry attempts.
-pub const DEFAULT_MAX_RETRIES: usize = 5;
+pub const DEFAULT_MAX_RETRIES: usize = 3;
 /// Default base delay between retries.
-pub const DEFAULT_RETRY_INTERVAL: Duration = Duration::from_secs(1);
+pub const DEFAULT_MIN_DELAY: Duration = Duration::from_secs(1);
```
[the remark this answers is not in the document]
LeoPatOZ: checked with david he seems fine with them!

0xNeshi (dismissed):

PR #141, opened 2025-10-28
fix: Retry updates

0xNeshi:
Towards #112

### DIFF
--- a/src/robust_provider.rs
+++ b/src/robust_provider.rs
@@ -95,7 +95,7 @@ impl<N: Network> RobustProvider<N> {
             error!(error = %e, "eth_getByBlockNumber failed");
         }
 
-        result?.ok_or(Error::BlockNotFound(number))
+        result?.ok_or_else(|| RobustProviderError::BlockNotFound(number))
     }
 
     /// Fetch the latest block number with retry and timeout.


src/robust_provider.rs
```diff
@@ -186,13 +173,10 @@ impl<N: Network> RobustProvider<N> {
...
     /// - Propagates any [`RpcError<TransportErrorKind>`] from the underlying retries.
-    async fn retry_with_total_timeout<T, F, Fut>(
-        &self,
-        operation: F,
-    ) -> Result<T, RobustProviderError>
+    async fn retry_with_total_timeout<T, F, Fut>(&self, operation: F) -> Result<T, Error>
     where
         F: Fn() -> Fut,
-        Fut: Future<Output = Result<T, RobustProviderError>>,
+        Fut: Future<Output = Result<T, RpcError<TransportErrorKind>>>,
```
0xNeshi: Reverting the change asked for in https://github.com/OpenZeppelin/Event-Scanner/pull/135#discussion_r2465619222.
Why? Since we know we're working only with `alloy`s providers, it is safe to specify the `RpcError` as the expected one for all operations.

PR #147, opened 2025-11-03
test: fix reorg_rescans_events_within_same_block

0xNeshi:
Towards #6

### DIFF
--- a/tests/live_mode/reorg.rs
+++ b/tests/live_mode/reorg.rs
@@ -18,7 +18,7 @@ use event_scanner::{
     robust_provider::RobustProvider,
 };
 
-#[test_log::test(tokio::test)]
+#[tokio::test]
 async fn reorg_rescans_events_within_same_block() -> anyhow::Result<()> {
     let (_anvil, provider, contract, filter) = setup_common(None, None).await?;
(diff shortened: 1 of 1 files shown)


src/robust_provider.rs
```diff
@@ -138,6 +138,10 @@ impl<N: Network> RobustProvider<N> {
         result
     }
 
+    pub async fn anvil_reorg(&self, options: ReorgOptions) -> TransportResult<()> {
+        self.provider.anvil_reorg(options).await
+    }
```
0xNeshi: `RobustProvider` should implement `AnvilApi`, so that it can be used throughout integration tests.

0xNeshi (commented):
Comments to be applied in the parent PR

src/robust_provider.rs
LeoPatOZ: I think this is why I originally exposed the underlying provider. Do you not think it makes more sense ? This way we don't cloud the robust provider with testing utilities

0xNeshi:
- clippy error should be addressed with https://github.com/OpenZeppelin/Event-Scanner/pull/147#discussion_r2485659975
- any other failing tests should be updated in parent PR, using this test update as reference

PR #157, opened 2025-11-05
Input validation

LeoPatOZ:
Resolves https://github.com/OpenZeppelin/Event-Scanner/issues/50

tldr:
- Adds validation to max_block_range (asserts > 0) on all scanners
- Asserts count in latest scanner > 0
- Validation to to / from block in historic mode. Neither can be bigger that latest block
- In order to achieve this more neatly we expose an shared internal build method. This way the api remains untouched and connect can be used for validation when needed.

### DIFF
--- a/src/event_scanner/scanner/mod.rs
+++ b/src/event_scanner/scanner/mod.rs
@@ -231,7 +231,7 @@ impl EventScannerBuilder<Unspecified> {
     /// # Example
     ///
     /// ```no_run
-    /// # use alloy::{network::Ethereum, primitives::Address, providers::{Provider, ProviderBuilder}};
+    /// # use alloy::{network::Ethereum, providers::{Provider, ProviderBuilder}};
     /// # use event_scanner::{EventFilter, EventScannerBuilder, Message, robust_provider::RobustProviderBuilder};
     /// # use tokio_stream::StreamExt;
     /// #


src/event_scanner/scanner/historic.rs
```diff
@@ -88,4 +131,36 @@ mod tests {
...
+            .from_block(latest_block + 100)
+            .to_block(latest_block)
+            .connect(provider)
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "Invalid historical block range")]
+    async fn test_to_block_above_latest_panics() {
```
0xNeshi: Add an additional test case when both are above latest

src/event_scanner/scanner/mod.rs
```diff
@@ -384,18 +393,15 @@ impl<M> EventScannerBuilder<M> {
     /// - Batch 4: blocks 1090–1099 (10 blocks)
     #[must_use]
     pub fn max_block_range(mut self, max_block_range: u64) -> Self {
+        assert!(max_block_range != 0, "max block range should be greater than zero");
```
0xNeshi: Should return error, not panic

src/event_scanner/scanner/mod.rs
```diff
@@ -322,6 +326,7 @@ impl EventScannerBuilder<Unspecified> {
     /// [reorg]: crate::ScannerStatus::ReorgDetected
     #[must_use]
     pub fn latest(count: usize) -> EventScannerBuilder<LatestEvents> {
+        assert!(count != 0, "count must be greater than 0");
```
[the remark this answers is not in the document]
LeoPatOZ: 1. 2. done!
 
 3. I think it makes to return an error. I personally think a count of 0 is semantically meaningless as it represents a misconfiguration or logic error in the caller's code. While the scanner won't break accepting it silently masks bugs.

src/error.rs
```diff
@@ -21,6 +21,15 @@ pub enum ScannerError {
 
     #[error("Operation timed out")]
     Timeout,
+
+    #[error("Block number {0} exceeds latest block {1}")]
+    BlockExceedsLatest(u64, u64),
```
0xNeshi:
```suggestion
    #[error("{0} {1} exceeds latest block {2}")]
    BlockExceedsLatest(&'static str, u64, u64),
```
Do you think it makes sense to include the field name that failed the validation? 
This would provide more context to the user :thinking:

0xNeshi (commented):
Close to merging!

src/event_scanner/scanner/historic.rs
```diff
@@ -18,6 +19,50 @@ impl EventScannerBuilder<Historic> {
...
+
+        let from_exceeds = from_num > latest_block;
+        let to_exceeds = to_num > latest_block;
+
+        match (from_exceeds, to_exceeds) {
+            (true, true) => Err(ScannerError::BlockExceedsLatest(
+                "from_block and to_block",
+                from_num.max(to_num),
+                latest_block,
+            ))?,
```
0xNeshi: latest = 4
from_block = 5
to_block = 6

resulting msg:
`from_block and to_block 6 exceeds the latest block 4`

Sounds kind of weird.

Let's follow the existing validation convention and just return the first validationerror we encounter; in this case, validate first just the `from_block` then if that passes, validate the `to_block`; emit errors only related to them.

PR #172, opened 2025-11-11
Add robust subscription

### LINKED ISSUE #167: feat: Scanner Live/Sync Modes Should Handle Subscriptions That Lag or Get Closed
Currently the scanner [converts the underlying Subscription into a non-erroring stream](https://github.com/OpenZeppelin/Event-Scanner/blob/a44fc87a4db5515de02134d2a134795f9149a99d/src/block_range_scanner.rs#L621). This is a problem, as it means that [errors get ignored](https://docs.rs/alloy-pubsub/latest/alloy_pubsub/struct.Subscription.html#method.into_stream). This would mean that the scanner will never try a fallback provider even if the currently used one hangs for too long (e.g. connection problems).

The scanner should perform the following:
- use the provider that initiated the subscription as longs as it works
- if the provider hangs for too long or the connection gets closed for whatever reason, the scanner should immediately switch to a fallback provider; it repeats this process for each provider it uses at the moment
- every X blocks, try to reconnect to the primary provider.

This _might_ include remove reliance on [SubscriptionStream](https://docs.rs/alloy-pubsub/latest/alloy_pubsub/struct.SubscriptionStream.html) (created via [subscription.into_stream](https://docs.rs/alloy-pubsub/latest/alloy_pubsub/struct.Subscription.html#method.into_stream) on [line 621](https://github.com/OpenZeppelin/Event-Scanner/blob/a44fc87a4db5515de02134d2a134795f9149a99d/src/block_range_scanner.rs#L621)) and using [Subscription::recv](https://docs.rs/alloy-pubsub/latest/alloy_pubsub/struct.Subscription.html#method.recv) directly.

### LINKED ISSUE #170: test: fix flaky `test_subscribe_fails_causes_backup_to_be_used`
[test](https://github.com/OpenZeppelin/Event-Scanner/blob/ab23884c6fe0fe0df6b98329075ac721fcf76818/src/robust_provider.rs#L560) 

Error: `channel closed`

https://github.com/OpenZeppelin/Event-Scanner/actions/runs/19225015176/job/54950193151

Problem seems to be that both Anvil nodes are started with `Anvil::new().try_spawn()?`, so maybe there's a conflict as the same instance is reused.

### LINKED ISSUE #185: `RobustProvider`: Client Should be Able to Pass Different Types of Provider-related Types Into `.fallback`
Currently we define the providers as a `Vec<P> where P: IntoProvider`.

The issue is once we pass in the 'main' provider, we now bind generic `P` to the primary provider's (sub)type.

So this is not possible:

```rust
        let ws_string: &str = "ws_example";
        let robust: RobustProvider<Ethereum> = RobustProviderBuilder::new(ws_provider)
            .fallback(ws_string)
            .call_timeout(Duration::from_secs(5))
            .build()
            .await?;
```

You need (where both primary and fallback are same type):

```rust
        let ws_string: &str = "ws_example";
        let robust: RobustProvider<Ethereum> = RobustProviderBuilder::new(ws_string)
            .fallback(ws_string)
            .call_timeout(Duration::from_secs(5))
            .build()
            .await?;
```

This is not the desired API. The client should have the benefit of 'mixing' provider types.

The solution to this might entail the use of `LazyProvider` wrapper types, so the internal vector type would be "fixed" to `Vec<LazyProvider<N>> where N: Network`, but internally the `LazyProvider` would be able to store different provider connection types.

Related #165

LeoPatOZ:
Resolves #167
Resolves #170
Resolves #185

### DIFF
--- a/src/robust_provider/subscription.rs
+++ b/src/robust_provider/subscription.rs
@@ -124,6 +124,14 @@ impl<N: Network> RobustSubscription<N> {
                         return Ok(header);
                     }
                     Err(recv_error) => {
+                        match recv_error {
+                            RecvError::Closed => {
+                                error!("Provider closed the subscription channel");
+                            }
+                            RecvError::Lagged(count) => {
+                                error!(skipped = count, "Receiver lagged");
+                            }
+                        }
                         return Err(recv_error.into());
                     }
                 },


src/block_range_scanner.rs
```diff
@@ -611,16 +611,29 @@ impl<N: Network> Service<N> {
...
+            Err(_) => false,
+        });
+
+        while let Some(result) = stream.next().await {
+            let incoming_block = match result {
+                Ok(block) => block,
+                Err(e) => {
+                    error!(error = %e, "Error receiving block from stream");
+                    // Error from subscription, exit the stream
+                    _ = sender.try_stream(e);
```
0xNeshi: don't forget to await

src/robust_provider.rs
```diff
@@ -177,6 +182,16 @@ impl<N: Network, P: IntoProvider<N>> RobustProviderBuilder<N, P> {
         self
     }
 
+    /// Set the timeout for subscription operations.
+    ///
+    /// This should be set higher than `max_timeout` to accommodate chains with slow block times.
+    /// Default is 2 minutes.
```
0xNeshi: Better to link to the constant directly, to avoid docs and code going out of sync

src/robust_provider.rs
```diff
@@ -329,26 +346,36 @@ impl<N: Network> RobustProvider<N> {
...
+
+        if let Err(e) = &subscription {
             error!(error = %e, "eth_subscribe failed");
+            return Err(e.clone());
         }
-        result
+
+        let subscription = subscription?;
+
+        Ok(RobustSubscription::new(subscription, self.clone(), DEFAULT_RECONNECT_INTERVAL))
```
0xNeshi: nit: rewrite as `match`

src/robust_provider.rs
```diff
@@ -390,13 +417,35 @@ impl<N: Network> RobustProvider<N> {
...
+        match self.try_fallback_providers(providers, &operation, require_pubsub, last_error).await {
+            Ok(value) => {
+                return Ok(value);
+            }
+            Err(e) => last_error = e,
+        }
+
+        // Return the last error encountered
+        error!("All providers failed or timed out - returning the last providers attempt's error");
+        Err(last_error)
```
0xNeshi: this is a more verbose way of saying: "_just return whatever `self.try_fallback_providers(providers, &operation, require_pubsub, last_error).await` returns_"

src/robust_subscription.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    /// Receive the next item from the subscription with automatic failover.
+    ///
+    /// This method will:
+    /// - Attempt to receive from the current subscription
+    /// - Handle errors by switching to fallback providers
+    /// - Periodically attempt to reconnect to the primary provider
+    /// - Will switch to fallback providers if subscription timeout is exhausted
```
0xNeshi: nit: for consistency use stars `*` instead of dashes `-`

0xNeshi (commented):
First pass

src/robust_provider.rs
LeoPatOZ: true

On another note should we rename max_timeout to max_rpc_timeout or something? as its really only specific to the an individual rpc http call as opposed to the subscription call.

src/robust_provider.rs
LeoPatOZ: is this better?

src/robust_provider.rs
LeoPatOZ: True - also did more refactors hopefully it makes the internal functions easier to reason about lmk

https://github.com/OpenZeppelin/Event-Scanner/pull/172/commits/18a6dd7ad7ce41c69c4f8832e11c059dfa0320dd

src/robust_provider.rs
0xNeshi: I see what you're getting at :thinking: Subscriptions are also RPC calls though. 

Maybe `call_timeout`?
(the `max_` is redundant, it's not like you can have a timeout smaller than this value)

src/robust_provider/provider.rs
```diff
@@ -484,4 +506,48 @@ mod tests {
...
+
+        while let Some(result) = stream.next().await {
+            let Ok(block) = result else {
+                break;
+            };
+
+            let block_number = block.number();
+
+            // At block 10, drop the primary provider to test failover
+            if block_number == 10 &&
```
0xNeshi: What is WS connection lags and it skips block 10? :P

More reliable to set up tests so that Anvil mines blocks on demand (`anvil_mine`) and the `anvil_1` is dropped outside the `while` loop.

src/robust_provider/provider.rs
```diff
@@ -484,4 +506,48 @@ mod tests {
...
+            .build()
+            .await?;
+
+        let subscription = robust.subscribe_blocks().await?;
+        let mut stream = subscription.into_stream();
+
+        while let Some(result) = stream.next().await {
+            let Ok(block) = result else {
+                break;
+            };
```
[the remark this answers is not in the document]
LeoPatOZ: removed

src/robust_provider/provider.rs
```diff
@@ -532,4 +546,127 @@ mod tests {
...
+        for _ in 0..30 {
+            ws_provider_2.anvil_mine(Some(10), None).await?;
+            let _ = stream.next().await.unwrap()?;
+            sleep(Duration::from_millis(100)).await;
+            // Mine on primary - should reconnect and receive from primary
+            ws_provider_1.anvil_mine(Some(1), None).await?;
+        }
+
+        let block = stream.next().await.unwrap()?;
+        assert_eq!(31 + 1, block.number());
```
0xNeshi: Shouldn't this be `30 + 1`?

- [30 times mined a single block on `ws_provider_1`](https://github.com/OpenZeppelin/Event-Scanner/pull/172/files/b47e062b33c63be4f62aa212e814aeb1f185e19a..4378c9ee0eb45176217eb9ac821a2b7c020f79dd#diff-ccfc5b365b4f693282550e2c47947b486c97ad2b05e625b40ff4550f50a69201R582) 
- [the original 1 block](https://github.com/OpenZeppelin/Event-Scanner/pull/172/files/b47e062b33c63be4f62aa212e814aeb1f185e19a..4378c9ee0eb45176217eb9ac821a2b7c020f79dd#diff-ccfc5b365b4f693282550e2c47947b486c97ad2b05e625b40ff4550f50a69201R571-R573)

src/robust_provider/provider.rs
```diff
@@ -532,4 +546,127 @@ mod tests {
...
+        let subscription = robust.subscribe_blocks().await?;
+
+        let mut stream = subscription.into_stream();
+
+        ws_provider_1.anvil_mine(Some(1), None).await?;
+        let block = stream.next().await.unwrap()?;
+        assert_eq!(1, block.number());
+
+        // simulate ws stream gone via no blocks mined > sub timeout
+        sleep(Duration::from_millis(600)).await;
```
0xNeshi: nit: for all tests - would a smaller `subscription_timeout` and shorter `sleep` work, too? E.g. `subscription_timeout = 100` and `sleep(200)`?

src/robust_provider/provider.rs
```diff
@@ -532,4 +546,127 @@ mod tests {
...
+        let robust = RobustProviderBuilder::fragile(ws_provider.clone())
+            .subscription_timeout(Duration::from_millis(500))
+            .build()
+            .await?;
+
+        let mut subscription = robust.subscribe_blocks().await?;
+
+        // simulate ws stream gone via no blocks mined > sub timeout
+        ws_provider.anvil_mine(Some(1), None).await?;
+        let _block = subscription.recv().await?;
```
0xNeshi:
```suggestion
        ws_provider.anvil_mine(Some(1), None).await?;
        let _ = subscription.recv().await?;
```
nit: forgot to delete comment?

0xNeshi (commented):
Getting closer to the complete feature

src/robust_provider/provider.rs
```diff
@@ -462,15 +468,26 @@ mod tests {
...
+            .subscription_timeout(Duration::from_secs(1))
             .build()
             .await?;
 
-        // force ws_provider to fail and return BackendGone
-        drop(anvil_1);
+        let mut subscription = robust.subscribe_blocks().await?;
+
+        ws_provider.anvil_mine(Some(1), None).await?;
+        assert_eq!(1, subscription.recv().await?.number());
```
0xNeshi: fix errors, e.g.:
https://github.com/OpenZeppelin/Event-Scanner/actions/runs/19424247503/job/55567945768?pr=172

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        let operation =
+            move |provider: RootProvider<N>| async move { provider.subscribe_blocks().await };
+
+        // Start searching from the next provider after the current one
+        let start_index = self.current_fallback_index.map_or(0, |idx| idx + 1);
+
+        let subscription = self
+            .robust_provider
+            .try_fallback_providers_from(&operation, true, last_error, start_index)
```
0xNeshi: circular dependency "smell": `provider.rs` depends on `subscription.rs`, and `subscription.rs` depends on `provider.rs`

we'll need to find a way to refactor that, if not in this PR then in a subsequent one

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// This method will:
+    /// * Attempt to receive from the current subscription
+    /// * Handle errors by switching to fallback providers
+    /// * Periodically attempt to reconnect to the primary provider
+    /// * Will switch to fallback providers if subscription timeout is exhausted
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if all providers have been exhausted and failed.
+    pub async fn try_recv_block(&mut self) -> Result<N::HeaderResponse, Error> {
```
0xNeshi: The prefix `try_` in the context of subscriptions implies an _immediate return_; verbalized: _"return the next block if any, otherwise immediately return an error"_

Rename this to just `recv` to align with alloy.

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+                        if let Err(err) = tx.send(Err(e)).await {
+                            warn!(error = %err, "Downstream channel closed, stopping stream");
+                        }
+                        break;
+                    }
+                }
+            }
+        });
+
+        RobustSubscriptionStream { inner: ReceiverStream::new(rx) }
```
0xNeshi: Instead of creating yet a separate channel that "eagerly" receives items and sends them to the robust stream, you can use `tokio-stream`'s [BroadcastStream](https://docs.rs/tokio-stream/latest/src/tokio_stream/wrappers/broadcast.rs.html#40-67) as inspiration, which "lazily" receives and streams new values. 
No additional channels necessary and no need for the inner `ReceiverStream`.

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+                    self.switch_to_fallback(error).await?;
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Try to reconnect to the primary provider if enough time has elapsed.
+    /// Returns true if reconnection was successful, false if it's not time yet or if it failed.
+    async fn try_reconnect_to_primary(&mut self) -> bool {
```
0xNeshi: nit: add `must_use`

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        let primary = self.robust_provider.primary();
+        let subscription =
+            self.robust_provider.try_provider_with_timeout(primary, &operation).await;
+
+        match subscription {
+            Ok(sub) => {
+                info!("Successfully reconnected to primary provider");
+                self.subscription = Some(sub);
+                self.current_fallback_index = None;
```
0xNeshi: Shouldn't `last_reconnect_attempt` be set to `None` here?

0xNeshi: re: https://github.com/OpenZeppelin/Event-Scanner/pull/172#discussion_r2533469544

Yes, but shouldn't it be set to `None`?

Correct me if I'm wrong, but currently if the primary provider is being used to get the next subscription item, on the following loop `try_reconnect_to_primary` is again called and if `last_attempt.elapsed() >= self.robust_provider.reconnect_interval`, it will reconnect to the already connected provider. 
By setting `last_reconnect_attempt` to `None` when primary provider is connected, this redundant operation would be skipped.

LeoPatOZ:
> These types of test failures should be handled with this PR, right? https://github.com/OpenZeppelin/Event-Scanner/actions/runs/19427639889/job/55578793510?pr=187

Hmm not exactly sure it would. Ive never seen this error before

src/robust_provider/mod.rs
```diff
@@ -2,8 +2,10 @@ pub mod builder;
 pub mod error;
 pub mod provider;
 pub mod provider_conversion;
+pub mod subscription;
```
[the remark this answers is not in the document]
LeoPatOZ: updated in merged PR

src/robust_provider/subscription.rs
LeoPatOZ: added in https://github.com/OpenZeppelin/Event-Scanner/pull/172/commits/736fe6ec3d640e82fe1077f8158b4f862e049536

src/robust_provider/subscription.rs
LeoPatOZ: fixed now!

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,254 @@
...
+                        error!(
+                            timeout_secs = subscription_timeout.as_secs(),
+                            "Subscription timeout - no block received, switching provider"
+                        );
+
+                        // If we're on a fallback, try reconnecting to primary one more time
+                        // before switching to the next fallback
+                        if self.current_fallback_index.is_some() &&
+                            self.try_reconnect_to_primary().await
+                        {
```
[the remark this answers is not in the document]
0xNeshi: I think ideally whenever FP1 fails, then PP fails again, the provider should try FP2; then if FP2 fails, try PP, then FP3.

"Reset" FPs only when one of them succeeds or if PP is retried and succeeds. So if FP2 succeeds, then after some time PP works again, once it fails, the provider starts trying from FP1 again.

Does that make sense?

Let me know if this turns out to be more complex to implement than anticipated.

src/robust_provider/mod.rs
```diff
@@ -1,3 +1,64 @@
+//! Robust, retrying wrapper around Alloy providers.
+//!
+//! This module exposes [`RobustProvider`], a small wrapper around Alloy's
+//! `RootProvider` that adds:
```
0xNeshi: add link to the actual docs/source, see how it's done in other docs

LeoPatOZ: Perhaps not sure what you mean.

I tried
`['RobustProvider'](crate::RobustProvider)` But it complains that they both link to same place so its redundant

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1213 @@
...
+        }
+        // No subscription available
+        Err(RpcError::Transport(TransportErrorKind::BackendGone).into())
+    }
+
+    /// Process subscription receive errors and handle failover
+    async fn process_recv_error(&mut self, recv_error: RecvError) -> Result<(), Error> {
+        match recv_error {
+            RecvError::Closed => {
+                error!("Subscription channel closed, switching provider");
```
0xNeshi:
```suggestion
                error!("Provider closed the subscription channel, switching provider");
```
nit: explicit about this being an external error

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1213 @@
...
+            .await;
+
+        match subscription {
+            Ok((sub, fallback_idx)) => {
+                self.subscription = Some(sub);
+                self.current_fallback_index = Some(fallback_idx);
+                Ok(())
+            }
+            Err(e) => {
+                error!(error = %e, "eth_subscribe failed - no fallbacks available");
```
0xNeshi: nit: redundant error log? 
Enough logs are emitted in `try_fallback_providers_from`. Maybe just refactor this to:
```rust
let (sub, fallback_idx) = subscription?;
self.subscription = ...
...
```

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1213 @@
...
+                Err(e)
+            }
+        }
+    }
+
+    /// Returns true if currently using a fallback provider
+    fn is_on_fallback(&self) -> bool {
+        self.current_fallback_index.is_some()
+    }
```
0xNeshi: Definitely weird that provider data is tracked in subscriptions

src/robust_provider/subscription.rs
```diff
@@ -57,13 +57,31 @@ impl<N: Network> RobustSubscription<N> {
...
+    ///    reconnect interval, reconnection will be delayed until the next block.
+    /// 2. **Forced reconnection**: Immediately when a fallback provider fails, before attempting
+    ///    the next fallback provider
+    ///
+    /// # Lag Handling
+    ///
+    /// If the subscription lags (receiver can't keep up with sender), the subscription tracks
+    /// consecutive lag occurrences. After 3 consecutive lags ([`MAX_LAG_COUNT`]), the subscription
+    /// will automatically switch to a fallback provider. The lag counter resets to 0 upon
+    /// successfully receiving a block.
```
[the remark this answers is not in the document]
LeoPatOZ: Okay I understand now.

I still have one more question is how to notify receiver that there was a lag. I think the correct approach is to send the Error::Lagged, then its up to the receiver to do something with that error or not

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+
+        // Start searching from the next provider after the current one
+        let start_index = self.current_fallback_index.map_or(0, |idx| idx + 1);
+
+        let subscription = self
+            .robust_provider
+            .try_fallback_providers_from(&operation, true, last_error, start_index)
+            .await;
+
+        let (sub, fallback_idx) = subscription?;
```
0xNeshi:
```suggestion
        let (sub, fallback_idx) = self
            .robust_provider
            .try_fallback_providers_from(&operation, true, last_error, start_index)
            .await?;
```

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+    for RobustSubscriptionStream<N>
+{
+    fn from(recv: RobustSubscription<N>) -> Self {
+        Self::new(recv)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
```
0xNeshi:
```suggestion
```

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+        sleep(SHORT_TIMEOUT + extra_delay + BUFFER_TIME).await;
+        assert_backend_gone_or_timeout(stream).await;
+    }
+
+    fn match_backend_gone_or_timeout(err: Error) {
+        match err {
+            Error::Timeout => {}
+            Error::RpcError(e) => {
+                assert!(
+                    matches!(e.as_ref(), RpcError::Transport(TransportErrorKind::BackendGone)),
```
0xNeshi: there's no more `BackendGone` error

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+    }
+
+    async fn assert_backend_gone_or_timeout(
+        stream: &mut RobustSubscriptionStream<alloy::network::Ethereum>,
+    ) {
+        let err = stream.next().await.unwrap().unwrap_err();
+        match_backend_gone_or_timeout(err);
+    }
+
+    async fn assert_backend_gone_or_timeout_after_delay(
```
0xNeshi: Same as previous

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+        let subscription = robust.subscribe_blocks().await?;
+        // Subscription is created successfully - is_empty() returns true initially (no pending
+        // messages)
+        assert!(subscription.is_empty());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_multiple_consecutive_recv_calls() -> anyhow::Result<()> {
```
0xNeshi: Update test: close the stream and assert `Error::Closed` is returned.

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+
+        Ok(())
+    }
+
+    // ----------------------------------------------------------------------------
+    // Basic Failover Tests
+    // ----------------------------------------------------------------------------
+
+    #[tokio::test]
+    async fn robust_subscription_stream_with_failover() -> anyhow::Result<()> {
```
0xNeshi: Add two more assertions to the test:
1. PP is not used after failover
2. Assert expected fields: `current_fallback_index`, `last_reconnect_attempt`

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1213 @@
...
+
+        let subscription = self
+            .robust_provider
+            .try_fallback_providers_from(&operation, true, last_error, start_index)
+            .await;
+
+        match subscription {
+            Ok((sub, fallback_idx)) => {
+                self.subscription = Some(sub);
+                self.current_fallback_index = Some(fallback_idx);
```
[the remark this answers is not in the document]
LeoPatOZ: Ah yes i forgot we cloned robust provider in subscription my bad 🤦  

Okay can create a tracking issue for this

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+    const RECONNECT_INTERVAL: Duration = Duration::from_millis(500);
+    const BUFFER_TIME: Duration = Duration::from_millis(100);
+
+    async fn spawn_ws_anvil() -> anyhow::Result<(AnvilInstance, RootProvider)> {
+        let anvil = Anvil::new().try_spawn()?;
+        let provider = ProviderBuilder::new().connect(anvil.ws_endpoint_url().as_str()).await?;
+        Ok((anvil, provider.root().to_owned()))
+    }
+
+    async fn assert_backend_gone_or_timeout(
```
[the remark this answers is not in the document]
LeoPatOZ: okay removed it, simpler this way

src/robust_provider/subscription.rs
LeoPatOZ: removed

src/robust_provider/subscription.rs
```diff
@@ -0,0 +1,1030 @@
...
+                    );
+
+                    self.switch_to_fallback(elapsed_err.into()).await?;
+                }
+            }
+        }
+    }
+
+    /// Process subscription receive errors
+    fn process_recv_error(recv_error: &RecvError) -> Result<(), Error> {
```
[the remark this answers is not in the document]
LeoPatOZ: okay updated

0xNeshi (dismissed):

0xNeshi (approved):

PR #187, opened 2025-11-17
feat: Update Chain Reorg Detection and Handling

### LINKED ISSUE #5: Chain Reorg Detection and Handling
### Description:
Implement blockchain reorg detection to ensure event ordering consistency and data integrity during chain reorgs.

Implementation:

- Detect when a reorg occurs (block hash changes)
- ~~Maintain recent block history (configurable depth, e.g., 12 blocks)~~ We should actually do:
    - Reorg to depth specified in config
    - Do 1 check of parent hash, if it matches return depth --> start streaming from there
    - If not, recursively find parent hash match
    - Return match --> stream from there 
- Re-process events from reorg point
- Ensure event ordering is maintained during reorg recovery
- Ensure reorg logic exists for both historical and live modes.

### LINKED ISSUE #56: Historic Mode Reorg Handling

### LINKED ISSUE #76: Live Mode Reorg Not Fully Robust
Our current subscription-based event scanning approach has a potential edge case where reorgs could go undetected if the subscription fails or skips blocks during the reorg period.
Edge Case Scenario
Timeline:
8 -> 9 -> 10 -> 11  (original chain)
8 -> 9' -> 10' -> 11'  (after reorg at block 9)

1. Block 8 processed successfully
2. Block 9 gets reorged to 9'
3. Subscription fails/skips and doesn't deliver block 9'
4. Subscription resumes and delivers block 10'
5. We have no way to detect that 10'  the latest block of a reorged chain
Current Detection Limitations
Why parent hash checks alone aren't sufficient:

We don't store previous block hashes, so we can't detect when a parent hash has changed
When we check block_10'.parent_hash == block_9'.hash, this will be true because we're comparing against the NEW hash of block 9'
We have no reference to the original block 9's hash to detect the discrepancy

## Root Cause
The fundamental issue is that we're not maintaining enough historical state to detect when the chain structure has changed between our last processed block and the current block.

## Possible Solution:

Finalized Block Chain Tracking

Maintain a ring buffer of block hashes from the latest finalized block to current head
On each new block, verify the entire chain of parent hashes back to our last known good state
Compare against stored chain to detect where reorg occurred

Image

Image

### LINKED ISSUE #155: feat: `ScannerStatus::SwitchingToLive` Should be Emitted Only When Scanner Catches Up with Latest Chain Tip
### Brief Sync Mode Logic Description

1. Get the current latest block number
2. Run **live mode** in the background and buffer all incoming blocks for later processing
3. Invoke **historic mode**, streaming all logs from `from_block` to `latest`
4. Stream `ScannerStatus::SwitchingToLive`
5. Start processing live mode blocks — i.e. stream events from buffered live blocks and all newly incoming blocks

---

### Example: Current Behavior

Assume the current chain blocks are:

```
1 - 2 - 3 - 4 - 5
```

Steps:

1. Scanner initiated with:

   ```rust
   sync::from_block(2)
   ```
2. The latest block is still `5`.
3. The scanner starts streaming logs from:

   ```
   2
   3
   ```
4. A new block `6` is added to the chain.
5. The scanner continues streaming:

   ```
   4
   ```
6. Another block `7` is added — the chain is now:

   ```
   ... - 5 - 6 - 7
   ```
7. The scanner streams:

   ```
   5
   SwitchingToLive
   6
   7
   ...
   ```

---

### Problem

At this point, the scanner is **not technically live** — it still has buffered blocks (`6` and `7`) to process before it’s actually synced with the latest chain tip.

---

### Expected Behavior

The `SwitchingToLive` notification should be emitted **only after** the scanner has processed all historic and buffered live blocks — i.e. once it reaches the latest chain tip.

The scanner should stream:

```
5 logs
6 logs
7 logs
```

If no new blocks have been added in the meantime, only then should it stream:

```
SwitchingToLive
```

---

### Main Challenge

Determine when the scanner has reached the **latest chain tip**, i.e. when it has processed:

* All buffered live blocks, and
* Is now processing only newly incoming live blocks.

0xNeshi:
Resolves #56 
Resolves #76 
Resolves #5
Resolves #155

### DIFF
--- a/src/block_range_scanner/common.rs
+++ b/src/block_range_scanner/common.rs
@@ -80,10 +80,11 @@ async fn get_first_block<
         match first_block {
             Ok(block) => return Some(block),
             Err(e) => {
-                error!(error = %e, "Error receiving block from stream");
                 match e {
                     subscription::Error::Lagged(_) => {
-                        continue;
+                        // scanner already accounts for skipped block numbers
+                        // next block will be the actual incoming block
+                        info!("Skipping Error::Lagged, next block should be the first live block");
                     }
                     subscription::Error::Timeout => {
                         _ = sender.try_stream(ScannerError::Timeout).await;
@@ -99,7 +100,7 @@ async fn get_first_block<
                     }
                 }
             }
-        };
+        }
     }
 
     None


src/block_range_scanner.rs
```diff
@@ -351,106 +373,19 @@ impl<N: Network> Service<N> {
...
-
-        Ok(())
+        let sync_handler = SyncHandler::new(
+            self.provider.clone(),
+            self.max_block_range,
+            start_id,
+            block_confirmations,
+            sender,
+        );
+        sync_handler.run().await
```
0xNeshi: this is a strategy-like pattern that we could use for mode handling. The logic is all contained within the strategy type.

src/block_range_scanner/common.rs
```diff
@@ -0,0 +1,334 @@
...
+
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn stream_live_blocks<N: Network>(
+    stream_start: BlockNumber,
+    subscription: Subscription<N::HeaderResponse>,
+    sender: &mpsc::Sender<BlockScannerResult>,
+    provider: &RobustProvider<N>,
+    block_confirmations: u64,
+    max_block_range: u64,
+    reorg_handler: &mut ReorgHandler<N>,
```
LeoPatOZ: nit: maybe we can initialise the reorg handle in this function?

src/block_range_scanner/reorg_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+    provider: RobustProvider<N>,
+    buffer: RingBuffer<BlockHash>,
+}
+
+impl<N: Network> ReorgHandler<N> {
+    pub fn new(provider: RobustProvider<N>) -> Self {
+        Self { provider, buffer: RingBuffer::new(10) }
+    }
+
+    pub async fn check(
```
LeoPatOZ: nit: Might help to put a comment above the function explaining how this works / edge cases. Im finding it hard to follow - especially at first glance trying to figure out what the return value is meant to represent

LeoPatOZ (commented):
Leaving these here for now, these are mostly very very initials thoughts just as i was reading through the PR (hence why they seem to be so disjunct).

Next couples of passes should hopefully be more focused on individual flows

src/block_range_scanner/common.rs
```diff
@@ -0,0 +1,334 @@
...
+    notify_after_first_block: bool,
```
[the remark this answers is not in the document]
LeoPatOZ: Maybe we extracted these functions out of the block range scanner too preemptively? 

I feel like they need so many of the same arguments that the block range scanner contains that it may as well be an impl of the connected scanner itself? Maybe in that case we have the reorg handler be a field of the block range scanner?

Im not exactly sure what the cleanest way to approach this is :( 

We can also think about separating the block range scanner by type just like the event scanner. We could have LiveBlockRangeScanner SyncFromBlockRange etc.. Perhaps it would make our lives easier 


Apologies Im just thinking out loud so im putting my thoughts here for now

src/block_range_scanner/common.rs
0xNeshi: We can't, because in sync mode we need to pass the reorg handler used in historic phase, which collects some blocks that could be used in live mode reorg handling

src/block_range_scanner/reorg_handler.rs
0xNeshi: Extracted the common ancestor calculation into a separate fn

src/block_range_scanner/common.rs
```diff
@@ -0,0 +1,334 @@
...
+        };
+
+        if !sender.try_stream(next_start_block..=batch_end_num).await {
+            return Some(batch_end);
+        }
+
+        batch_count += 1;
+        if batch_count % 10 == 0 {
+            debug!(batch_count = batch_count, "Processed historical batches");
+        }
```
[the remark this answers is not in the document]
0xNeshi: I don't want to intentionally introduce unrelated changes to the PR. If you have an idea where this could be emitted instead, feel free to open a small PR

src/block_range_scanner/common.rs
```diff
@@ -0,0 +1,334 @@
...
+            Ok(block) => block,
+            Err(e) => {
+                error!(batch_start = next_start_block, batch_end = batch_end_num, error = %e, "Failed to get ending block of the current batch");
+                _ = sender.try_stream(e).await;
+                return None;
+            }
+        };
+
+        if !sender.try_stream(next_start_block..=batch_end_num).await {
+            return Some(batch_end);
```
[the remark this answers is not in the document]
0xNeshi: probably leftover from some old logic, removing

LeoPatOZ (approved):

PR #201, opened 2025-11-21
sync::from_latest and latest Modes Should Return Notification::NoPastLogsFound Instead Of Empty Vec

0xNeshi:
Towards #50 - added missing from/to block validation to `latest` mode

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,7 +35,7 @@ jobs:
           egress-policy: audit
 
       - name: Fetch Repository
-        uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3
 
       - name: Install stable toolchain
         uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1.15.2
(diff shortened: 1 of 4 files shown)


src/event_scanner/scanner/common.rs
```diff
@@ -120,12 +120,16 @@ pub fn spawn_log_consumers<N: Network>(
...
+                } else {
+                    info!(count = collected.len(), "Logs found");
                     collected.reverse(); // restore chronological order
-                }
 
-                info!("Sending collected logs to consumer");
-                _ = sender.try_stream(collected).await;
+                    info!("Sending collected logs to consumer");
+                    _ = sender.try_stream(collected).await;
+                }
```
LeoPatOZ: nit:
```suggestion
                if collected.is_empty() {
                    info!("No logs found");
                    _ = sender.try_stream(Notification::NoPastLogsFound).await;
                    break;
                }
                 
                    info!(count = collected.len(), "Logs found");
collected.reverse(); // restore chronological order

                    info!("Sending collected logs to consumer");
                    _ = sender.try_stream(collected).await;
                }
```

src/types.rs
```diff
@@ -5,16 +5,32 @@ use tracing::{info, warn};
 
 use crate::ScannerError;
 
-#[derive(Debug, Clone)]
+/// Messages streamed by the scanner to subscribers.
+///
+/// Each message represents either data (logs), an error, or a notification about the scanner's
```
LeoPatOZ: nit:
```suggestion
/// Each message represents either data or a notification about the scanner's
```
No more error :D 

Also not always logs?

LeoPatOZ (dismissed):
🚀

src/event_scanner/scanner/common.rs
0xNeshi: No loop to break out of :smile: Will apply the formatting though

PR #211, opened 2025-11-25
Allow for multiple provider types

### LINKED ISSUE #185: `RobustProvider`: Client Should be Able to Pass Different Types of Provider-related Types Into `.fallback`

LeoPatOZ:
Resolves #185

### DIFF
--- a/src/robust_provider/provider.rs
+++ b/src/robust_provider/provider.rs
@@ -3,7 +3,7 @@ use std::{fmt::Debug, time::Duration};
 use alloy::{
     eips::{BlockId, BlockNumberOrTag},
     network::{Ethereum, Network},
-    primitives::BlockHash,
+    primitives::{BlockHash, BlockNumber},
     providers::{Provider, RootProvider},
     rpc::types::{Filter, Log},
     transports::{RpcError, TransportErrorKind},
@@ -89,7 +89,7 @@ impl<N: Network> RobustProvider<N> {
     /// # Errors
     ///
     /// See [retry errors](#retry-errors).
-    pub async fn get_block_number(&self) -> Result<u64, Error> {
+    pub async fn get_block_number(&self) -> Result<BlockNumber, Error> {
         info!("eth_getBlockNumber called");
         let result = self
             .try_operation_with_failover(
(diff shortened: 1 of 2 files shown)


src/robust_provider/provider_conversion.rs
```diff
@@ -82,8 +82,14 @@ pub trait IntoRobustProvider<N: Network = Ethereum> {
...
     async fn into_robust_provider(self) -> Result<RobustProvider<N>, Error> {
         RobustProviderBuilder::new(self).build().await
     }
 }
+
+pub(crate) async fn into_root_provider<N: Network, P: IntoProvider<N>>(
+    provider: P,
+) -> Result<RootProvider<N>, Error> {
+    Ok(provider.into_provider().await?.root().to_owned())
+}
```
0xNeshi: Implement this in `trait IntoProvider` directly

LeoPatOZ: Okay actually updated the into_provider to be into_root_provider. As fundamentally the provider needs to be able to be converted to a root provider.

PR #212, opened 2025-11-27
Add robust subscription: tests

0xNeshi:
Towards #167


LeoPatOZ:
Okay if you dont i might take the idea of these tests and add them directly to the robust provider. Might be easier to port them one by one then trying to fix everything in this PR

PR #221, opened 2025-12-04
Optimization: Fetching Historic Block Ranges Before the Finalized Block Should Omit Reorg Check

### LINKED ISSUE #102: Optimization: Fetching Historic Block Ranges Before the Finalized Block Should Omit Reorg Check

0xNeshi:
Resolves #102

### DIFF
--- a/src/block_range_scanner/common.rs
+++ b/src/block_range_scanner/common.rs
@@ -389,7 +389,7 @@ pub(crate) async fn stream_historical_range<N: Network>(
     Some(())
 }
 
-/// Assumes that `min_block <= next_start_block <= end`.
+/// Assumes that `min_common_ancestor <= next_start_block <= end`, performs no internal checks.
 pub(crate) async fn stream_range_with_reorg_handling<N: Network>(
     min_common_ancestor: BlockNumber,
     mut next_start_block: BlockNumber,


src/block_range_scanner/common.rs
```diff
@@ -319,9 +326,72 @@ struct LiveStreamingState<N: Network> {
...
+            _ = sender.try_stream(e).await;
+            return None;
+        }
+    };
+
+    // no reorg check for finalized blocks
+    let mut batch_start = start;
+    let finalized_batch_end = finalized.min(end);
+    while batch_start < finalized_batch_end {
+        let batch_end = batch_start.saturating_add(max_block_range - 1).min(end);
```
LeoPatOZ: I think this may be a bug. If end > finalized the final batch may send unfinalized blocks.

I believe it should be this:

```suggestion
        let batch_end = batch_start.saturating_add(max_block_range - 1).min(finalized_batch_end);
```

0xNeshi: There is a way to allow the scanner to stream non-finalized blocks in this "phase" as well, for two reasons:
- it would be the last stream in this phase
- a reorg check could be set immediately after in case a non-finalized block was streamed in this phase.

But now that I think about it, it's currently impossible to test this behavior. Better to apply your suggestion for now, as it's simpler to reason about.

0xNeshi: The more I think about it, the more I'm in favor of your suggestion, simply because it's simpler, and avoids unnecessary edge cases

PR #224, opened 2025-12-04
Replaced reorg detection with parent hash verification

### LINKED ISSUE #214: Refactor reorg detection to use ring buffer parent hash verification
## Summary
Replace the current reorg detection approach (checking if last streamed block hash exists on-chain) with a ring buffer approach that verifies incoming blocks against their parent hashes. This is cleaner to reason about and more performant in common scenarios.

## Current Approach
Check whether the last streamed block's hash still exists on-chain. If not, walk back until we find a matching hash, then re-stream from that point.
## Proposed Approach
Store block hashes and numbers in a ring buffer as they arrive. For each incoming block, verify its parent hash against the buffer. On mismatch, walk back to find the fork point.
### Scenario 1: Stream correctly rewinds to fork point

`1 → 2 → 3 → REORG → 2`

| | Current | Proposed |
|---|---|---|
| RPC calls | 3 | 0 |
| End buffer state | `1` (var holds `3`) | `1, 2, 3` |

**Current**: Check block 3 hash on-chain (missing), check block 2 (missing), check block 1 (match). Stream `2..=3`, fetch block 3 for future checks.

**Proposed**: Block 2 arrives, parent hash matches stored block 1 hash. Stream block 2, update buffer, continue.

### Scenario 2: Stream returns a block past the fork point

`1 → 2 → 3 → (2 reorged) → 3`

| | Current | Proposed |
|---|---|---|
| RPC calls | 3 | 1 |
| End buffer state | `1` (var holds `3`) | `1, 2, 3` |

**Current**: Block 3 arrives, check stored block 3 hash on-chain (missing), check block 2 (missing), check block 1 (match). Stream `2..=3`, fetch block 3.

**Proposed**: Block 3 arrives, parent hash doesn't match stored block 2 hash. Fetch block 2, its parent matches block 1. Stream `2 → 3`, update buffer.

### Scenario 3: Stream skips a range with reorg mid-gap

`1 → 2 → 3 → (6 reorged) → 8`

| | Current | Proposed |
|---|---|---|
| RPC calls | 1 | 5 |
| End buffer state | `1, 2, 3` (var holds `8`) | `1, 2, ..., 8` |

**Current**: Check stored block 3 hash on-chain (exists). No reorg detected. Stream `4 → 8`.

**Proposed**: Fetch hashes for blocks 4–7, verify chain consistency i.e. 4 parents matches 3 and 8 parents matches 7. Stream `4 → 8`, update buffer. The scanner will process blocks 4 → 8, which already reflect the post-reorg canonical chain

##  Edge Cases to Consider

- **Reorg deeper than buffer size** : Fetch finalised block and send that range instead.

##  Conclusion
The proposed approach is cleaner and more performant for the common cases (scenarios 1 and 2). Scenario 3 requires more RPC calls but maintains fuller buffer state which may be beneficial for subsequent reorg detection.

yug49:
Resolves #214 

### Description
Refactors reorg detection from on-chain hash existence checks to parent hash verification against a ring buffer.

### Changes
- **`ring_buffer.rs`**: Added `BlockInfo<H>` struct storing block number + hash, with `find_by_number()` and `truncate_from()` methods
- **`reorg_handler.rs`**: Replaced hash existence checks with parent hash verification; added `handle_gap()` for intermediate block fetching

### Scenarios
| Scenario | Before | After |
|----------|--------|-------|
| Direct reorg (1→2→3→REORG→2) | 3 RPC calls | 0 RPC calls |
| Block past fork (1→2→3→(2 reorged)→3) | 3 RPC calls | 1 RPC call |
| Gap with reorg (1→2→3→(6 reorged)→8) | 1 RPC call | 5 RPC calls (but maintains fuller buffer) |

### Edge cases handled
- Empty buffer (first block)
- Reorg deeper than buffer size → fallback to finalized block
- Gaps in block numbers → fetch intermediate blocks

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -129,7 +129,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Check spelling of files in the workspace
-        uses: crate-ci/typos@9066e9940a8a05b98fb4733c62a726f83c9e57f8 # v1.43.3
+        uses: crate-ci/typos@78bc6fb2c0d734235d57a2d6b9de923cc325ebdd # v1.43.4
         
   hack:
     # cargo-hack checks combinations of feature flags to ensure that features are all additive
@@ -149,7 +149,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1.15.2
 
       - name: cargo install cargo-hack
-        uses: taiki-e/install-action@509565405a8a987e73cf742e26b26dcc72c4b01a # v2.67.26
+        uses: taiki-e/install-action@288875dd3d64326724fa6d9593062d9f8ba0b131 # v2.67.30
         with:
           tool: cargo-hack
 
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -52,6 +52,6 @@ jobs:
           path: results.sarif
           retention-days: 5
       - name: Upload SARIF to GitHub Code Scanning
-        uses: github/codeql-action/upload-sarif@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2  # v4.32.2
+        uses: github/codeql-action/upload-sarif@9e907b5e64f6b83e7804b09294d44122997950d6  # v4.32.3
         with:
           sarif_file: results.sarif
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Install cargo-nextest
         if: steps.cache-cargo-nextest.outputs.cache-hit != 'true'
-        uses: taiki-e/install-action@509565405a8a987e73cf742e26b26dcc72c4b01a # v2.67.26
+        uses: taiki-e/install-action@288875dd3d64326724fa6d9593062d9f8ba0b131 # v2.67.30
         with:
           tool: cargo-nextest
 


src/block_range_scanner/ring_buffer.rs
```diff
@@ -81,10 +87,32 @@ impl<T> RingBuffer<T> {
...
+    ///
+    /// Performs a reverse linear search since we typically look for recent blocks.
+    pub fn find_by_number(&self, number: BlockNumber) -> Option<&BlockInfo<H>> {
+        self.inner.iter().rev().find(|info| info.number == number)
+    }
+
+    /// Truncates the buffer to remove all blocks after (and including) the given block number.
+    /// This is used when a reorg is detected to remove the reorged blocks.
+    pub fn truncate_from(&mut self, from_number: BlockNumber) {
+        self.inner.retain(|info| info.number < from_number);
```
LeoPatOZ: I believe doing this is slightly more efficient 

```rust
let idx = self.inner.partition_point(|info| info.number < from_number);
self.inner.truncate(idx);
```

retain visit every element runs the predicate on each and for elements that stay it swaps them into position. However since we are in sorted order this should be better since we just dropping the tail

src/block_range_scanner/reorg_handler.rs
```diff
@@ -51,85 +56,96 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
...
+            info!(block_number = block_number, "Buffer empty, adding first block");
+            self.buffer.push(BlockInfo { number: block_number, hash: block_hash });
+            return Ok(None);
         }
 
-        // return last finalized block as common ancestor
+        // Check if this is a previously seen block (duplicate check)
+        // This happens when checking `previous_batch_end` to verify it wasn't reorged
+        if let Some(buffered) = self.buffer.find_by_number(block_number) {
+            if buffered.hash == block_hash {
```
LeoPatOZ: I believe this is wrong / may never happen. 

We are working under the assumption that if a new block comes with the same number it will have a different hash i..e the only way this happened is if a reorg occurred.


The buffer would be:

`[100, 101, 102]`

If block 102 comes in its hash should not exist in the buffer. In that case we'll need to do a parent hash i..e the normal flow of the reorg check. 


If this is true we can perhaps remove verify_block_on_chain fn and this if branch.

@0xNeshi lmk if you agree

PR #232, opened 2025-12-06
feat: configurable maximum stream capacity

### LINKED ISSUE #228: Max Stream Capacity Should be Configurable
Currently we set a predefined `MAX_BUFFERED_MESSAGES` constant:
https://github.com/OpenZeppelin/Event-Scanner/blob/cdcc2cc56f210af06d6728f6ea78ff9197f338d3/src/block_range_scanner.rs#L595

Maybe some devs would prefer to set this themselves.

yug49:
Resolves #228 

## Description

This PR makes the hardcoded `MAX_BUFFERED_MESSAGES` constant configurable via a new `max_stream_capacity` builder method, allowing users to tune the internal channel buffer sizes based on their specific use case. The `MAX_BUFFERED_MESSAGES` constant (50,000) was previously hardcoded and used for both `mpsc::channel` and `broadcast::channel` capacities. 

## Changes

- Added `max_stream_capacity` field to `BlockRangeScanner`, `ConnectedBlockRangeScanner`, and `BlockRangeScannerClient`
- Added `max_stream_capacity()` builder method to `EventScannerBuilder` for all scanner modes
- Updated `handle_stream()` to accept configurable capacity parameter
- All internal channel creations now use the configurable capacity
- Default value remains `50,000` for backward compatibility

### DIFF
--- a/src/block_range_scanner.rs
+++ b/src/block_range_scanner.rs
@@ -395,14 +395,17 @@ impl<N: Network> Service<N> {
             _ => (start_block_num, end_block_num),
         };
 
-        info!(start_block = start_block_num, end_block = end_block_num, "Syncing historical data");
+        info!(
+            start_block = start_block_num,
+            end_block = end_block_num,
+            "Normalized the block range"
+        );
 
         tokio::spawn(async move {
             let mut reorg_handler =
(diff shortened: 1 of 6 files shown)


LeoPatOZ:
Resolved in:
https://github.com/OpenZeppelin/Event-Scanner/pull/266

Issue #236, opened 2025-12-08
Investigate potential flaky test test_to_block_above_latest_returns_error and test_subscription_lagged_error


LeoPatOZ:
For some reason these two tests have failed locally. If this happens in CI we need to investigate why / fix this - for now I cant reliably get them to  fail.

LeoPatOZ:
https://github.com/OpenZeppelin/Event-Scanner/actions/runs/20105660143/job/57688672692?pr=240 

Example of test_subscription_lagged_error failing

PR #239, opened 2025-12-08
test: fixed flaky test `test_subscription_lagged_error`

yug49:
Hey!

I noticed these two tests were flagged as potentially flaky in #236 , so I took a closer look.

**What I found:**

1. ~~**`test_to_block_above_latest_returns_error`** (in both latest.rs and historic.rs) - There is a small race window between calling `get_block_number()` and `connect()`. If a block happens to be mined in that window, the `latest` value in the error won't match what we captured, causing the assertion to fail. I relaxed the assertion to just verify `latest < max` instead of exact equality. This way we still confirm the error is correct, just without being too strict about timing.~~ edit: no longer included in the PR

2. **`test_subscription_lagged_error`** - The test was mining blocks immediately after creating the subscription, but the subscription might not be fully established yet. Added a small delay (`BUFFER_TIME`) before mining to give it time to set up.

I ran each tests locally and they passed consistently, but would love your thoughts on whether this approach makes sense or if you would prefer a different fix.

### DIFF
--- a/src/event_scanner/scanner/latest.rs
+++ b/src/event_scanner/scanner/latest.rs
@@ -371,8 +371,7 @@ mod tests {
         match result {
             Err(ScannerError::BlockExceedsLatest("to_block", max, latest)) => {
                 assert_eq!(max, latest_block + 100);
-                // We only verify latest <= max, not exact equality with latest_block,as a block could be mined between get_block_number() and connect().
-                assert!(latest < max);
+                assert_eq!(latest, latest_block);
             }
             _ => panic!("Expected BlockExceedsLatest error"),
         }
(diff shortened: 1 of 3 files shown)


src/event_scanner/scanner/latest.rs
```diff
@@ -339,7 +339,8 @@ mod tests {
         match result {
             Err(ScannerError::BlockExceedsLatest("to_block", max, latest)) => {
                 assert_eq!(max, latest_block + 100);
-                assert_eq!(latest, latest_block);
+                // We only verify latest <= max, not exact equality with latest_block,as a block could be mined between get_block_number() and connect().
+                assert!(latest < max);
```
0xNeshi: Same as above.

0xNeshi (commented):
@yug49 we got confirmation for what the cause of the subscription error is:
https://github.com/OpenZeppelin/Event-Scanner/issues/236#issuecomment-3640973275

Please apply https://github.com/OpenZeppelin/Event-Scanner/pull/239/files#r2601359860, remove other changes, and we can merge

0xNeshi (approved):
Thanks for the contribution!

PR #240, opened 2025-12-08
feat: maintain proper log ordering in rewind reorg detection + tests

LeoPatOZ:


src/event_scanner/scanner/common.rs
```diff
@@ -155,25 +155,31 @@ pub fn spawn_log_consumers_in_collection_mode<N: Network>(
             // Only used for CollectLatest
             let mut collected = Vec::with_capacity(count);
 
+            // Tracks common ancestor block during reorg recovery for proper log ordering
+            let mut reorg_ancestor: Option<u64> = None;
+
```
0xNeshi: This whole module is really hard to unit test. We don't support mocking any of the externalities, e.g. provider, or just the helper functions... 
We'll need to work on that.

src/event_scanner/scanner/common.rs
```diff
@@ -252,6 +262,26 @@ pub fn spawn_log_consumers_in_collection_mode<N: Network>(
...
 }
 
+/// Collects logs into the buffer, either prepending (reorg recovery) or appending (normal).
+/// Returns `true` if collection is complete (reached count limit).
+fn collect_logs(collected: &mut Vec<Log>, logs: Vec<Log>, count: usize, prepend: bool) -> bool {
+    if prepend {
+        let new_logs: Vec<_> = logs.into_iter().rev().take(count).collect();
+        let keep = count.saturating_sub(new_logs.len());
+        collected.truncate(keep);
+        collected.splice(..0, new_logs);
```
0xNeshi: Doesn't this misbehave if the reorg is deep enough that it takes processing at least 2 block ranges to fully handle it?
Why? When a reorg is detected in Rewind, it currently re-streams reorged ranges in reverse `tip -> common ancestor`, which would make `collect_logs` push newer logs first.

LeoPatOZ: hmm perhaps im wrong but it believe it sends the range from common -> tip?

Wouldnt the stream output something like this:

1. 90..=100        ← pre reorg
2. ReorgDetected(95)
3. 96..=100        ← roerg range 
4. 80..=89         ← continue rewind
5. 70..=79
6. 60..=69
7. 50..=59

?

0xNeshi: In your example, what would the block range stream be if `ReorgDetected(85)`?

LeoPatOZ: I guess it depends when it happens but lets say:

90..=100 
80..=89 
70..=79
ReorgDetected(85)
96..=100 ← reorg range
86..=95 ← reorg range
60..=69 ← continue rewind
50..=59

0xNeshi (approved):

Issue #52, opened 2025-09-22
Improve Logging Logic


0xNeshi:
Logging the processes in our service will drastically improve maintainability, for both us and package users, especially because the service is multi-threaded. 

Investigate exact crate to rely on for logging, maybe first investigate [tracing](https://docs.rs/tracing/latest/tracing/).

We got some feedback on logs, and they generally currently seem to be annoying, as they just clutter the output.

Quote:
> Based on my experience, I feel like lots of logs are redundant and i always set RUST_LOG=info,event_scanner=warn :smile: . Since I have my own logs each time there is a message / notification. And lots of logs from scanner which is INFO level is like should be DEBUG / TRACE level for me, e.g. the current block ranges to scan, the current raw logs scanned...
> 
> And i have app level logs for each message / notificaiton / error, so probly what i care is, if there is anything unexpected.
> 
> For current INFO level logs, i do feel it's redundant like the block ranges stuffs. And if there is a lot of raw logs for one shot, the log size will be huge that is annoying

Some general guidelines that we need to discuss and agree upon:
- the point of logs is:
   - notify the dev what the scanner is doing
   - notify the dev what data the scanner is processing
   - create history of events leading up to an error/bug
- write meaningful log entries
- use appropriate log levels:
   - error - unexpected failure that needs to be debugged and fixed
   - warning/info - should probably not be used, as users generally set their own logs
   - debug/trace - should use this the most
       - debug - log vars, choices made (e.g. in an `if` statement)
       - trace - general info (e.g. "hey, I'm here now", "I'm calling this function", "the call succeeded!" etc.)
- logs should indicate where they originated (block range scanner, event scanner, maybe even the exact function?)
- (optional) logs should contain some sort of trace ID, to enable tracking the chain of events and possibly indicating in which thread/task logs are being emitted

PR #258, opened 2025-12-15
feat: clean up tracing in robust provider

LeoPatOZ:
Towards #52 

For now deals with robust provider logging.

Removes majority of `info!` adds debug tracing to rpc methods (similar to how alloy does it).

### DIFF
--- a/src/robust_provider/subscription.rs
+++ b/src/robust_provider/subscription.rs
@@ -166,6 +166,7 @@ impl<N: Network> RobustSubscription<N> {
         }
     }
 
+    #[cfg_attr(feature = "tracing", tracing::instrument(level = "trace", skip(self)))]
     async fn switch_to_fallback(&mut self, last_error: CoreError) -> Result<(), Error> {
         // If we're on a fallback, try primary first before moving to next fallback
         if self.is_on_fallback() && self.try_reconnect_to_primary(true).await {


src/robust_provider/provider.rs
```diff
@@ -111,20 +111,17 @@ impl<N: Network> RobustProvider<N> {
     /// # Errors
     ///
     /// See [retry errors](#retry-errors).
+    #[instrument(level = "debug", skip(self))]
     pub async fn get_block_by_number(
```
0xNeshi: Thinking about this some more, what's really the use case from a user's perspective for logging internal function calls, even in debug mode? Are they really going to debug our scanner using their logs?

src/robust_provider/provider.rs
```diff
@@ -361,31 +333,33 @@ impl<N: Network> RobustProvider<N> {
...
         for (fallback_idx, provider) in fallback_providers {
             if require_pubsub && !Self::supports_pubsub(provider) {
-                info!("Fallback provider {} doesn't support pubsub, skipping", fallback_idx + 1);
+                warn!(
+                    provider_num = fallback_idx + 1,
+                    "Fallback provider doesn't support pubsub, skipping"
+                );
                 continue;
             }
-            info!("Attempting fallback provider {}/{}", fallback_idx + 1, num_fallbacks);
```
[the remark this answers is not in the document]
LeoPatOZ: Okay added back the trace and then added the instrument attribute

src/robust_provider/provider.rs
0xNeshi: > Perhaps we should remove them

I think so, too

src/robust_provider/provider.rs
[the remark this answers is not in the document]
0xNeshi: Oh, ok then!

src/robust_provider/provider.rs
```diff
@@ -111,20 +111,17 @@ impl<N: Network> RobustProvider<N> {
     /// # Errors
     ///
     /// See [retry errors](#retry-errors).
+    #[instrument(level = "debug", skip(self))]
```
0xNeshi:
```suggestion
    #[instrument(level = "tracing", skip(self))]
```
Let's support the lowest level for all `instrument` attributes, we don't know how devs will invoke this

src/robust_provider/subscription.rs
```diff
@@ -163,7 +147,7 @@ impl<N: Network> RobustSubscription<N> {
             return false;
         }
 
-        info!("Attempting to reconnect to primary provider");
+        trace!("Attempting to reconnect to primary provider");
```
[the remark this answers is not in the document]
LeoPatOZ: updated (to instrument)

src/robust_provider/provider.rs
```diff
@@ -384,34 +351,41 @@ impl<N: Network> RobustProvider<N> {
...
                 );
                 continue;
             }
-            info!(fallback_index = fallback_idx + 1, "Attempting fallback provider");
+
+            tracing::trace!(
+                fallback_provider_index = fallback_idx + 1,
+                total_num_fallbacks = num_fallbacks,
+                "Attempting fallback provider"
+            );
```
0xNeshi:
```suggestion
           trace!(
                fallback_provider_index = fallback_idx + 1,
                total_num_fallbacks = num_fallbacks,
                "Attempting fallback provider"
            );
```
this is a regular feature-gated trace

src/robust_provider/subscription.rs
```diff
@@ -162,7 +148,7 @@ impl<N: Network> RobustSubscription<N> {
...
```
0xNeshi: Redundant, since the function is instrumented

src/robust_provider/provider.rs
```diff
@@ -357,6 +322,7 @@ impl<N: Network> RobustProvider<N> {
             .await
     }
 
+    #[instrument(level = "trace", skip(self, operation, last_error))]
     pub(crate) async fn try_fallback_providers<T: Debug, F, Fut>(
```
0xNeshi: do we need this instrumentation when it just invokes the instrumented `try_fallback_providers_from`?

src/robust_provider/provider.rs
```diff
@@ -372,6 +338,7 @@ impl<N: Network> RobustProvider<N> {
             .map(|(value, _idx)| value)
     }
 
+    #[instrument(level = "trace", skip(self, operation, last_error))]
     pub(crate) async fn try_fallback_providers_from<T: Debug, F, Fut>(
```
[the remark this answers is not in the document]
LeoPatOZ: done!

PR #261, opened 2025-12-16
Docs improvements

### LINKED ISSUE #256: Improve Documentation
- Document all `pub` and unintuitive `pub(crate)`: fields, enums, functions, traits, structs
    - (optional) document all `pub(crate)` types
- Document intricacies of scanner, edge cases etc.

### LINKED ISSUE #260: Receiving `RecvError::Lagged` in `EventScanner` Causes Skipped Block Ranges
Relevant file:
https://github.com/OpenZeppelin/Event-Scanner/blob/e3abf2d2881548675c4c56e5b05e4edb7d10b897/src/event_scanner/scanner/common.rs#L156

The scanner currently only logs that a certain number of ranges for skipped, and carries on without ever processing the skipped ranges.
This means that if a scanner lags, the caller will never receive logs from the skipped ranges, and they will not even be aware that this had happened.

As this error is by definition caused on the caller side (they do not process the messages on time), the most logical (and straightforward) way is to just pass `RecvError::Lagged` to the caller so they can handle it.

0xNeshi:
Resolves #256 
Resolves #260

https://github.com/OpenZeppelin/Event-Scanner/pull/266 somehow reverted #268 , so this PR adds it back...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,8 +47,11 @@
 //! retry and fail over across multiple RPC endpoints.
 //!
 //! [finalized]: alloy::eips::BlockNumberOrTag::Finalized
-pub mod block_range_scanner;
 
+#[macro_use]
+mod logging;
+
+pub mod block_range_scanner;
 pub mod robust_provider;
 #[cfg(any(test, feature = "test-utils"))]
 pub mod test_utils;
--- a/src/robust_provider/provider.rs
+++ b/src/robust_provider/provider.rs
@@ -12,7 +12,6 @@ use backon::{ExponentialBuilder, Retryable};
 use futures::TryFutureExt;
 use thiserror::Error;
 use tokio::time::{error as TokioError, timeout};
-use tracing::{error, info};
 
 use crate::robust_provider::RobustSubscription;
 
@@ -229,7 +228,7 @@ impl<N: Network> RobustProvider<N> {
     /// * [`Error::Timeout`] - if the overall operation timeout elapses (i.e. exceeds
     ///   `call_timeout`).
     pub async fn get_latest_confirmed(&self, confirmations: u64) -> Result<u64, Error> {
-        info!("get_latest_confirmed called with confirmations={}", confirmations);
+        info!(configurations = confirmations, "get_latest_confirmed called");
         let latest_block = self.get_block_number().await?;
         let confirmed_block = latest_block.saturating_sub(confirmations);
         Ok(confirmed_block)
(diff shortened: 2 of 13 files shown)


src/event_scanner/scanner/sync/from_block.rs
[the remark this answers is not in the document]
0xNeshi: Applied both + listed specific errors that can be returned.

src/event_scanner/scanner/mod.rs
```diff
@@ -24,34 +57,61 @@ mod sync;
 /// Default number of maximum concurrent fetches for each scanner mode.
 pub const DEFAULT_MAX_CONCURRENT_FETCHES: usize = 24;
 
+/// Typestate marker indicating that a scanner mode has not been selected yet.
 #[derive(Default)]
 pub struct Unspecified;
+
+/// Mode marker for historical range scanning.
+///
+/// Constructed via [`EventScannerBuilder::historic`].
```
LeoPatOZ: I think here would be a good place to write what each mode does. So that when you click on Historic its explicity says 

'Scans a specified block range before the latest block etc...,

Image

Wdyt? 

For now theres no obvious place where it tells you what each mode does. (just from clicking around the doc.rs site)

0xNeshi: `EventScannerBuilder::historic` has what you're looking for.

Maybe the Mode marker docs should state:
```rust
For more details on this scanner mode, see `EventScannerBuilder::historic`.
```
?

0xNeshi: https://github.com/OpenZeppelin/Event-Scanner/pull/261/commits/f3a3197476b4ea575b657570d729f5a168859483

LeoPatOZ: nice yeah that works

src/lib.rs
```diff
@@ -1,3 +1,52 @@
+//! Event-Scanner is a library made to stream EVM event logs.
```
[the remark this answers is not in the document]
LeoPatOZ: hmm dont have anything specific should thought we could add that in for 'free' since we have the read me already. But yeah maybe we omit it for now.

src/robust_provider/provider.rs
LeoPatOZ: Just noticed we link error to error defined in `try_operation_with_failover` however this functions visibility is set to crate which means it doesnt show up in the docs so the link doesnt go anywhere.

We can either

1. set this function visibility to pub - maybe this is actually useful that the robust provider exposes this method, this allows anyone using the robust provider to call any custom RPC method with retry even if our provider doesnt directly support it yet.
2. We move the errors description to the pub enum errors and link it there?

src/lib.rs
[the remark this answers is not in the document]
0xNeshi: You'll notice in the project you linked to that it references a _crate level_ README, not the _project level_ README. 

What you suggested would make sense if the event scanner was physically located (figuratively speaking) in a subdirectory of our workspace, not in the root; this is because it would then have its own README, into which we could copy/paste the current _src/lib.rs_ mod-level docs, and in the file do `#![doc = include_str!(".././README.md")]`.
This is a (minor) project tree design issue more than anything else.

Given that the main README could expand unpredictably, I'm in favor of leaving everything as-is for now.

src/robust_provider/provider.rs
0xNeshi: I like 1). 

Given the facts that:
- `try_operation_with_failover` returns `CoreError` while provider functions return `Error`
- some provider functions return `BlockNotFound`, which does not exist as a variant of `CoreError` 

I think that just linking to `try_operation_with_failover` in the error sections does not make sense anymore.

Made `try_operation_with_failover` and `CoreError` public (https://github.com/OpenZeppelin/Event-Scanner/pull/261/commits/4f6cc3d47a0d09dc275c19d267ef9c0fa7a92303) and listed explicit errors in each provider function's error doc (https://github.com/OpenZeppelin/Event-Scanner/pull/261/commits/4f6cc3d47a0d09dc275c19d267ef9c0fa7a92303)

@LeoPatOZ wdyt?

LeoPatOZ (dismissed):
Sweet thanks for doing this

LeoPatOZ (approved):

PR #265, opened 2025-12-16
feat: use Mode instead of unspecified

LeoPatOZ:
Not sure if its worth adding but just so type M can be more explicit in Event Scanner.  Then 

Screenshot 2025-12-16 at 17 26 21

So we can have this:

Screenshot 2025-12-16 at 17 29 40


0xNeshi (commented):
Much simpler is to update the generic parameter name from `M` to `Mode`:
```rust
pub struct EventScanner<Mode = Unspecified, N: Network = Ethereum> {
    config: Mode,
    block_range_scanner: ConnectedBlockRangeScanner<N>,
    listeners: Vec<EventListener>,
}
```

image

PR #266, opened 2025-12-17
Make max Stream Capacity Configurable

### LINKED ISSUE #228: Max Stream Capacity Should be Configurable

0xNeshi:
Resolves #228 

⚠️ THIS PR WAS MOSTLY AI GENERATED

### DIFF
--- a/src/block_range_scanner.rs
+++ b/src/block_range_scanner.rs
@@ -93,7 +93,7 @@ pub const DEFAULT_MAX_BLOCK_RANGE: u64 = 1000;
 // copied form https://github.com/taikoxyz/taiko-mono/blob/f4b3a0e830e42e2fee54829326389709dd422098/packages/taiko-client/pkg/chain_iterator/block_batch_iterator.go#L19
 pub const DEFAULT_BLOCK_CONFIRMATIONS: u64 = 0;
 
-pub const DEFAULT_BUFFER_CAPACITY: usize = 50000;
+pub const DEFAULT_STREAM_BUFFER_CAPACITY: usize = 50000;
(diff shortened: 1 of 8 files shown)


src/event_scanner/scanner/mod.rs
LeoPatOZ: Dont we need to validate buffer > 0 it here? + add custom Error?

src/block_range_scanner.rs
```diff
@@ -166,11 +174,18 @@ impl BlockRangeScanner {
         self,
         provider: impl IntoRobustProvider<N>,
     ) -> Result<ConnectedBlockRangeScanner<N>, ScannerError> {
+        if self.max_block_range == 0 {
+            return Err(ScannerError::InvalidMaxBlockRange);
+        }
+        if self.buffer_capacity == 0 {
+            return Err(ScannerError::InvalidBufferCapacity);
```
0xNeshi: both of these should really be validated in BRS

LeoPatOZ: yeah i remember having this discussion but we decided to leave validation to the event scanner 

Edit: but its fine we have them here now

LeoPatOZ (dismissed):

LeoPatOZ (approved):

PR #270, opened 2025-12-18
Use macro wrappers for feature-gated `tracing` macros, turning all of them "off" by default

0xNeshi:
Towards #52 

Replaces #267

### DIFF
--- a/src/robust_provider/provider.rs
+++ b/src/robust_provider/provider.rs
@@ -9,6 +9,7 @@ use alloy::{
     transports::{RpcError, TransportErrorKind},
 };
 use backon::{ExponentialBuilder, Retryable};
+use futures::TryFutureExt;
 use thiserror::Error;
 use tokio::time::{error as TokioError, timeout};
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
LeoPatOZ: I think we can mark this as optional = true?

Then in https://github.com/OpenZeppelin/Event-Scanner/blob/5ccbc340f9d312e30e8eb27679244d16e44723c2/Cargo.toml#L77

We can do:

tracing = ["dep:tracing"]

I tried this but things werent working due to workspace dependencies or something. 

Do you think its worth adding this?

LeoPatOZ: feature "robust_tracing" 😆

src/logging.rs
LeoPatOZ: Do you not think its confusing to have the same macro name?

Its now not immediately obvious which traces are on by default and which are hidden behind our feature flag (could be more error prone)

LeoPatOZ (approved):

PR #274, opened 2025-12-18
Refactor brs to remove command based design

LeoPatOZ:
This is the first of multiple block range scanner refactors


0xNeshi (approved):
Excellent 😍

PR #277, opened 2025-12-19
Make reorg traces on by default

### LINKED ISSUE #52: Improve Logging Logic

LeoPatOZ:
Resolves #52

### DIFF
--- a/src/block_range_scanner.rs
+++ b/src/block_range_scanner.rs
@@ -10,7 +10,7 @@
 //!
 //! # Output stream
 //!
-//! Streams returned by [`BlockRangeScannerClient`] yield [`BlockScannerResult`] items:
+//! Streams returned by [`ConnectedBlockRangeScanner`] yield [`BlockScannerResult`] items:
 //!
 //! - `Ok(ScannerMessage::Data(range))` for a block range to process.
 //! - `Ok(ScannerMessage::Notification(_))` for scanner notifications.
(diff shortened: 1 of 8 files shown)


0xNeshi:
Closing in favor of https://github.com/OpenZeppelin/Event-Scanner/pull/281

PR #294, opened 2026-01-07
Optimise cloning of event filter

LeoPatOZ:


src/event_scanner/block_range_handler.rs
```diff
@@ -75,13 +75,19 @@ impl<N: Network> StreamHandler<N> {
...
+            broadcast_channel_capacity,
+        }
     }
 
     fn spawn(&self, range_tx: &Sender<BlockScannerResult>) -> JoinSet<()> {
         let mut join_set = JoinSet::new();
 
-        for listener in self.listeners.iter().cloned() {
+        for idx in 0..self.listeners.len() {
+            let listeners = Arc::clone(&self.listeners);
```
0xNeshi: Hmm, it might completely unnecessary to clone anything really :thinking: We could update `spawn` and `handle` signatures to take owned `self`, and here iterate over owned `listeners`.

```rust
fn spawn(self, ...) ... {
    let mut join_set = JoinSet::new();
    
    for listener in self.listeners {
        ...
        // tokio task "consumes" the owned listener, so no cloning is necessary
    }
}
```
Wdyt?

PR #302, opened 2026-01-14
Use 0 as earliest block in finalized block edge case

### LINKED ISSUE #301: Investigate if falling back to earliest block is appropriate for finalised block edge case
The problem is described in this issue #298  and was fixed in https://github.com/OpenZeppelin/Event-Scanner/pull/300. 

The idea of the PR is to default to the earliest block if finalised block is not available. The question remains, should we default to the earliest block (requiring an RPC call) or set the default block number to 0. 

Core Issue: Investigate whether 0 is **always** the earliest block in a chain, if so we should update the [`fetch_finalized_or_earliest_block_number` function](https://github.com/OpenZeppelin/Event-Scanner/blob/release-v1.0.0/src/block_range_scanner/common.rs#L54) to return 0 instead of earliest.

LeoPatOZ:
Resolves #301 

Effectively just turns on reorg checking for all block ranges.

### DIFF
--- a/src/block_range_scanner/common.rs
+++ b/src/block_range_scanner/common.rs
@@ -393,9 +393,9 @@ pub(crate) async fn stream_historical_range<N: Network>(
     // NOTE: Edge case - If the chain is too young to expose finalized blocks (height < finalized
     // depth) just use zero.
     // Since we use the finalized block number only to determine whether to run reorg checks
-    // or not, this is a "low-stakes" RPC call, for which, for simplicity, we can default to `0` even
-    // on errors. Here `0` is used because it effectively just enables reorg checks.
-    // If there was actually a provider problem, any subsequent provider call will catch and 
+    // or not, this is a "low-stakes" RPC call, for which, for simplicity, we can default to `0`
+    // even on errors. Here `0` is used because it effectively just enables reorg checks.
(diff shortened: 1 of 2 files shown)


src/block_range_scanner/common.rs
```diff
@@ -412,10 +390,13 @@ pub(crate) async fn stream_historical_range<N: Network>(
     provider: &RobustProvider<N>,
     reorg_handler: &mut ReorgHandler<N>,
 ) -> Option<()> {
-    let finalized = fetch_finalized_or_earliest_block_number(provider, sender).await?;
+    // NOTE: Edge case - If the chain is too young to expose finalized blocks (height < finalized
+    // depth) just use zero
+    let finalized_block_num =
```
0xNeshi:
```suggestion
    // NOTE: Edge case - If the chain is too young to expose finalized blocks (height < finalized
    // depth) just use zero.
    // Since we use the finalized block number only to determine whether to run reorg checks
    // or not, this is a "low-stakes" RPC call, for which, for simplicity, we can default to `0` even
    // on errors. Here `0` is used because it effectively just enables reorg checks.
    // If there was actually a provider problem, any subsequent provider call will catch and 
    // properly log it and return the error to the caller.
    let finalized_block_num =
```

0xNeshi (approved):

PR #352, opened 2026-05-25
fix: detect reorged sidechain blocks by canonical hash

YoGhurt111:
## Summary
- Detect reorgs by comparing a previously scanned block against the current canonical block at the same height.
- Require buffered common-ancestor candidates to still be canonical before accepting them.
- Add a regression test for providers that keep reorged sidechain blocks queryable by hash.

## Root Cause
Some execution clients can still return a reorged sidechain block from `eth_getBlockByHash` after the canonical block at the same height has changed. The previous reorg check treated a successful hash lookup as proof that the buffered/scanned block was still canonical, so same-height replacements could be missed and the affected range would not be replayed.

## Testing
- `cargo test --features test-utils detects_reorg_when_old_hash_is_available_but_not_canonical`
- `cargo check --features test-utils`

Note: `cargo fmt --check` on this machine reports unrelated formatting diffs in existing files (`src/block_range_scanner/common.rs` and `src/event_scanner/block_range_handler.rs`) due to the stable rustfmt handling of the repository rustfmt config. Those unrelated files were not changed in this PR.

### DIFF
--- a/src/block_range_scanner/reorg_handler.rs
+++ b/src/block_range_scanner/reorg_handler.rs
@@ -29,6 +29,58 @@ pub trait ReorgHandler<N: Network> {
     ) -> Result<Option<N::BlockResponse>, ScannerError>;
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloy::{
+        consensus::Header as ConsensusHeader,
+        network::Ethereum,
+        providers::{RootProvider, mock::Asserter},
+        rpc::{client::RpcClient, types::Block},
+    };
+    use robust_provider::RobustProviderBuilder;
+
+    fn block(number: u64, hash: BlockHash) -> Block {
+        Block::empty(alloy::rpc::types::Header {
+            hash,
+            inner: ConsensusHeader { number, ..Default::default() },
+            total_difficulty: None,
+            size: None,
+        })
+    }
+
+    #[tokio::test]
+    async fn detects_reorg_when_old_hash_is_available_but_not_canonical() -> anyhow::Result<()> {
+        let parent_hash = BlockHash::repeat_byte(0x09);
+        let old_hash = BlockHash::repeat_byte(0x0a);
+        let new_hash = BlockHash::repeat_byte(0x0b);
+        let parent_block = block(9, parent_hash);
+        let old_block = block(10, old_hash);
+        let new_block = block(10, new_hash);
+
+        let asserter = Asserter::new();
+        asserter.push_success(&Some(new_block.clone()));
+        asserter.push_success(&Some(old_block.clone()));
+        asserter.push_success(&Some(new_block));
+        asserter.push_success(&Some(parent_block.clone()));
+        asserter.push_success(&Some(parent_block.clone()));
+        asserter.push_success(&Some(parent_block));
+
+        let provider = RootProvider::<Ethereum>::new(RpcClient::mocked(asserter));
+        let provider = RobustProviderBuilder::fragile(provider).build().await?;
+        let mut handler = DefaultReorgHandler::new(provider, RingBufferCapacity::Limited(10));
+        handler.buffer.push(parent_hash);
+        handler.buffer.push(old_hash);
+
+        let common_ancestor = handler.check(&old_block).await?.expect("reorg should be detected");
+
+        assert_eq!(common_ancestor.header().number(), 9);
+        assert_eq!(common_ancestor.header().hash(), parent_hash);
+
+        Ok(())
+    }
+}
+
 /// Default implementation of [`ReorgHandler`] that uses an RPC provider.
 #[derive(Clone, Debug)]
 pub(crate) struct DefaultReorgHandler<N: Network = Ethereum> {
@@ -103,13 +155,20 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
         while let Some(&block_hash) = self.buffer.back() {
             trace!(block_hash = %block_hash, "Checking if buffered block exists on chain");
             match self.provider.get_block_by_hash(block_hash).await {
-                Ok(common_ancestor) => {
+                Ok(candidate) if self.is_canonical(candidate.header()).await? => {
                     debug!(
                         common_ancestor_hash = %block_hash,
-                        common_ancestor_number = common_ancestor.header().number(),
+                        common_ancestor_number = candidate.header().number(),
                         "Found common ancestor"
                     );
-                    return self.return_common_ancestor(common_ancestor).await;
+                    return self.return_common_ancestor(candidate).await;
+                }
+                Ok(_candidate) => {
+                    trace!(
+                        block_hash = %block_hash,
+                        "Buffered block is no longer canonical, removing from buffer"
+                    );
+                    _ = self.buffer.pop_back();
                 }
                 Err(robust_provider::Error::BlockNotFound) => {
                     // block was reorged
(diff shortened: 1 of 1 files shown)


0xNeshi:
Confirmed the finding that Geth can return a non-canonical block by hash, if the block is still in the node's DB.

Sources:
- [rawdb docs](https://pkg.go.dev/github.com/ethereum/go-ethereum/core/rawdb#ReadAllHashes) state it stores both canonical and non-canonical blocks
- [eth_getBlockByNumber](https://etclabscore.github.io/core-geth/JSON-RPC-API/modules/eth/#eth_getblockbynumber) is explicit about returning only canonical blocks at the specified height, while [eth_getBlockByHash](https://etclabscore.github.io/core-geth/JSON-RPC-API/modules/eth/#eth_getblockbyhash) isn't.


@YoGhurt111 feel free to list other clients that do the same

src/block_range_scanner/reorg_handler.rs
```diff
@@ -103,13 +155,20 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
...
                     debug!(
                         common_ancestor_hash = %block_hash,
-                        common_ancestor_number = common_ancestor.header().number(),
+                        common_ancestor_number = candidate.header().number(),
                         "Found common ancestor"
                     );
-                    return self.return_common_ancestor(common_ancestor).await;
+                    return self.return_common_ancestor(candidate).await;
+                }
+                Ok(_candidate) => {
```
0xNeshi: Fold into the `Err` arm below:
```rust
Ok(_) | Err(robust_provider::Error::BlockNotFound) => {
...
```

src/block_range_scanner/reorg_handler.rs
```diff
@@ -103,13 +155,20 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
         while let Some(&block_hash) = self.buffer.back() {
             trace!(block_hash = %block_hash, "Checking if buffered block exists on chain");
             match self.provider.get_block_by_hash(block_hash).await {
-                Ok(common_ancestor) => {
+                Ok(candidate) if self.is_canonical(candidate.header()).await? => {
```
0xNeshi: Consider that we can perform the canonical check with a single RPC call if we make the following changes:
1. Update [buffer: RingBuffer<BlockHash>](https://github.com/OpenZeppelin/Event-Scanner/pull/352/changes#diff-6e72c0781dac31ce15f7505199eaaac08e0293540e404b1d25db85977338e1d5L36) to something like `buffer: RingBuffer<(BlockNumber, BlockHash)>`, so that we store both the checked block hash and number
2. Update the [match](https://github.com/OpenZeppelin/Event-Scanner/pull/352/changes#diff-6e72c0781dac31ce15f7505199eaaac08e0293540e404b1d25db85977338e1d5L105) here to something like:
```rust
        while let Some(&(candidate_number, candidate_hash)) = self.buffer.back() {
            match self.provider.get_block_by_number(candidate_number).await {
                Ok(canonical) if canonical.header().hash() == candidate_hash => {
```
3. Remove `is_canonical`

src/block_range_scanner/reorg_handler.rs
```diff
@@ -29,6 +29,58 @@ pub trait ReorgHandler<N: Network> {
     ) -> Result<Option<N::BlockResponse>, ScannerError>;
 }
 
+#[cfg(test)]
+mod tests {
```
0xNeshi: Move tests module to the bottom of the file

YoGhurt111: Fixed in [6d9b26a](https://github.com/OpenZeppelin/Event-Scanner/pull/352/commits/6d9b26a22d44e37f1fa4ee6963b0ef7215c30b45)

src/block_range_scanner/reorg_handler.rs
YoGhurt111: Fixed in [6d9b26a](https://github.com/OpenZeppelin/Event-Scanner/pull/352/commits/6d9b26a22d44e37f1fa4ee6963b0ef7215c30b45)

src/block_range_scanner/reorg_handler.rs
YoGhurt111: Fixed in [6d9b26a](https://github.com/OpenZeppelin/Event-Scanner/pull/352/commits/6d9b26a22d44e37f1fa4ee6963b0ef7215c30b45)

src/block_range_scanner/reorg_handler.rs
0xNeshi:
```suggestion
    primitives::{BlockHash, BlockNumber},
```

src/block_range_scanner/reorg_handler.rs
```diff
@@ -33,7 +33,7 @@ pub trait ReorgHandler<N: Network> {
 #[derive(Clone, Debug)]
 pub(crate) struct DefaultReorgHandler<N: Network = Ethereum> {
     provider: RobustProvider<N>,
-    buffer: RingBuffer<BlockHash>,
+    buffer: RingBuffer<(u64, BlockHash)>,
```
0xNeshi:
```suggestion
    buffer: RingBuffer<(BlockNumber, BlockHash)>,
```
nit: clearer what is being stored

src/block_range_scanner/reorg_handler.rs
```diff
@@ -100,20 +101,27 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
...
-            match self.provider.get_block_by_hash(block_hash).await {
-                Ok(common_ancestor) => {
+        while let Some(&(candidate_number, candidate_hash)) = self.buffer.back() {
+            trace!(
+                block_number = candidate_number,
+                block_hash = %candidate_hash,
+                "Checking if buffered block is canonical"
+            );
+            match self.provider.get_block_by_number(candidate_number.into()).await {
+                Ok(candidate) if candidate.header().hash() == candidate_hash => {
```
0xNeshi:
```suggestion
                Ok(canonical) if canonical.header().hash() == candidate_hash => {
```
nit: differentiate between the stored block at the given height (candidate) and the actual block at the given height (canonical)

src/block_range_scanner/reorg_handler.rs
```diff
@@ -100,20 +101,27 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
...
                     debug!(
-                        common_ancestor_hash = %block_hash,
-                        common_ancestor_number = common_ancestor.header().number(),
+                        common_ancestor_hash = %candidate_hash,
+                        common_ancestor_number = candidate.header().number(),
```
0xNeshi:
```suggestion
                        common_ancestor_number = candidate_number,
```

src/block_range_scanner/reorg_handler.rs
```diff
@@ -100,20 +101,27 @@ impl<N: Network> ReorgHandler<N> for DefaultReorgHandler<N> {
...
                         "Found common ancestor"
                     );
-                    return self.return_common_ancestor(common_ancestor).await;
+                    return self.return_common_ancestor(candidate).await;
```
0xNeshi:
```suggestion
                    return self.return_common_ancestor(canonical).await;
```

0xNeshi (approved):
Looks good, thank you!

0xNeshi:
We have a signed commit requirement, so closing this PR with unsigned commits in favor of https://github.com/OpenZeppelin/Event-Scanner/pull/354

PR #354, opened 2026-05-29
fix: detect reorged sidechain blocks by canonical hash

YoGhurt111:
## Summary
- Detect reorgs by comparing a previously scanned block against the current canonical block at the same height.
- Store buffered block numbers alongside hashes so common ancestor lookup uses a single canonical block-by-number request.
- Add regression coverage for providers that keep reorged sidechain blocks queryable by hash.

## Root Cause
Some execution clients can still return a reorged sidechain block from `eth_getBlockByHash` after the canonical block at the same height has changed. Treating a successful hash lookup as proof that a buffered/scanned block is still canonical can miss same-height replacements, so the affected range is not replayed.

## Fix
This change compares the saved block hash with the canonical block returned by `eth_getBlockByNumber` at the saved height. The reorg buffer stores `(BlockNumber, BlockHash)`, allowing common ancestor detection to validate canonicality with one RPC call per buffered candidate.

## Testing
- `cargo test --features test-utils detects_reorg_when_old_hash_is_available_but_not_canonical`
- `cargo check --features test-utils`

This PR replaces #352 with a single signed commit.

### DIFF
--- a/src/block_range_scanner/reorg_handler.rs
+++ b/src/block_range_scanner/reorg_handler.rs
@@ -2,7 +2,7 @@ use alloy::{
     consensus::BlockHeader,
     eips::BlockNumberOrTag,
     network::{BlockResponse, Ethereum, Network, primitives::HeaderResponse},
-    primitives::BlockHash,
+    primitives::{BlockHash, BlockNumber},
 };
 use robust_provider::RobustProvider;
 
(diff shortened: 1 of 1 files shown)


0xNeshi (approved):
Excellent, thanks !

0xNeshi:
Replaces https://github.com/OpenZeppelin/Event-Scanner/pull/352

